PR #49, opened 2023-10-02
.github/workflows: Remove test check from e2e

kfirtoledo:
Remove tests/k8s.sh from the pr-check of e2e-test.


kfirtoledo:
I see it works now, I will continue to watch before I remove it

PR #54, opened 2023-10-02
[New]: Add heartbeat for the new controlplane

kfirtoledo:
Add heartbeat mechanism to new controlplane for checking liveness between peers

### DIFF
--- a/pkg/controlplane/client.go
+++ b/pkg/controlplane/client.go
@@ -3,8 +3,11 @@ package controlplane
 import (
 	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
+	"sync"
+	"time"
 
 	"github.com/sirupsen/logrus"
 
@@ -13,12 +16,21 @@ import (
 	"github.com/clusterlink-net/clusterlink/pkg/util/jsonapi"
 )
 
+const (
+	// heartbeatInterval is the time lapse between consecutive heartbeat requests to a responding peer.
+	heartbeatInterval = 10 * time.Second
+	// heartbeatRetransmissionTime is the time lapse between consecutive heartbeat requests to a non-responding peer.
+	heartbeatRetransmissionTime = 60 * time.Second
+)
+
 // client for accessing a remote peer.
 type client struct {
 	// jsonapi clients for connecting to the remote peer (one per each gateway)
-	clients []*jsonapi.Client
-
-	logger *logrus.Entry
+	clients  []*jsonapi.Client
+	lastSeen time.Time
+	active   bool
+	lock     sync.RWMutex
+	logger   *logrus.Entry
 }
 
 // remoteServerAuthorizationResponse represents an authorization response received from a remote controlplane server.
@@ -78,16 +90,91 @@ func (c *client) Authorize(req *api.AuthorizationRequest) (*remoteServerAuthoriz
 	return resp, nil
 }
 
+// IsActive returns if the peer is active or not.
+func (c *client) IsActive() bool {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	return c.active
+}
+
+// setActive the peer status (active or not).
+func (c *client) setActive(active bool) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.active = active
+	if active || c.lastSeen.IsZero() {
+		c.lastSeen = time.Now()
+	}
+}
+
+// GetHeartbeat get a heartbeat from other peers.
+func (c *client) getHeartbeat() error {
+	var retErr error
+	// copy peer clients array aside
+	peerClients := make([]*jsonapi.Client, len(c.clients))
+	{
+		c.lock.RLock()
+		defer c.lock.RUnlock()
+		copy(peerClients, c.clients)
+	}
+
+	for _, client := range peerClients {
+		serverResp, err := client.Get(api.HeartbeatPath)
+		if err != nil {
+			retErr = errors.Join(retErr, err)
+			continue
+		}
+
+		if serverResp.Status == http.StatusOK {
+			return nil
+		}
+
+		retErr = errors.Join(retErr, fmt.Errorf("unable to get heartbeat (%d), server returned: %s",
+			serverResp.Status, serverResp.Body))
+	}
+
+	return retErr // Return an error if all client targets are unreachable
+}
+
+// heartbeatMonitor checks all peers for responsiveness, every fixed amount of time.
+func (c *client) heartbeatMonitor() {
+	c.logger.Info("Start sending heartbeat requests to peer")
+	ticker := time.NewTicker(heartbeatInterval)
+	defer ticker.Stop()
+	for {
+		t := time.Now()
+		if c.IsActive() || (!c.IsActive() && (t.Sub(c.lastSeen) > heartbeatRetransmissionTime)) ||
+			c.lastSeen.IsZero() {
+			if err := c.getHeartbeat(); err != nil {
+				if c.IsActive() {
+					c.logger.Errorf("Unable to get heartbeat from peer  Error: %v", err.Error())
+					c.setActive(false)
+				}
+			} else {
+				c.setActive(true)
+			}
+		}
+
+		// wait till it's time for next heartbeat round
+		<-ticker.C
+	}
+}
+
 // newClient returns a new Peer API client.
 func newClient(peer *store.Peer, tlsConfig *tls.Config) *client {
 	clients := make([]*jsonapi.Client, len(peer.Gateways))
 	for i, endpoint := range peer.Gateways {
 		clients[i] = jsonapi.NewClient(endpoint.Host, endpoint.Port, tlsConfig)
 	}
-	return &client{
-		clients: clients,
+	c := &client{
+		clients:  clients,
+		active:   false,
+		lastSeen: time.Time{},
 		logger: logrus.WithFields(logrus.Fields{
 			"component": "peer-client",
 			"peer":      peer}),
 	}
+
+	go c.heartbeatMonitor()
+	return c
 }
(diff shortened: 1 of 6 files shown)


elevran (changes requested):

pkg/controlplane/client.go
```diff
@@ -78,16 +89,82 @@ func (c *client) Authorize(req *api.AuthorizationRequest) (*remoteServerAuthoriz
...
+func (c *client) getHeartbeat() error {
+	var retErr error
+	// copy peer clients array aside
+	peerClients := make([]*jsonapi.Client, len(c.clients))
+	c.lock.RLock()
+	copy(peerClients, c.clients)
+	c.lock.RUnlock()
+
+	for _, client := range peerClients {
+		serverResp, err := client.Post(api.HeartbeatPath, nil)
```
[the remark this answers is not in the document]
kfirtoledo: Done

pkg/controlplane/client.go
```diff
@@ -78,16 +89,82 @@ func (c *client) Authorize(req *api.AuthorizationRequest) (*remoteServerAuthoriz
 	return resp, nil
 }
 
+// Active returns if the peer is active or not.
+func (c *client) Active() bool {
+	c.lock.RLock()
+	a := c.active
+	c.lock.RUnlock()
```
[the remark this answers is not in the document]
kfirtoledo: Done,
In the next function, I don't want to wait until the end of the function to release the RUlock.

pkg/controlplane/client.go
```diff
@@ -78,16 +89,82 @@ func (c *client) Authorize(req *api.AuthorizationRequest) (*remoteServerAuthoriz
...
+	// copy peer clients array aside
+	peerClients := make([]*jsonapi.Client, len(c.clients))
+	c.lock.RLock()
+	copy(peerClients, c.clients)
+	c.lock.RUnlock()
+
+	for _, client := range peerClients {
+		serverResp, err := client.Post(api.HeartbeatPath, nil)
+		if err != nil {
+			retErr = err
```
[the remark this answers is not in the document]
kfirtoledo: The problem is that we can have several addresses to the same peer, so I return Err and not active only if all of them are inactive.

pkg/controlplane/client.go
```diff
@@ -78,16 +89,82 @@ func (c *client) Authorize(req *api.AuthorizationRequest) (*remoteServerAuthoriz
...
+		} else {
```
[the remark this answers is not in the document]
kfirtoledo: see comment above

pkg/controlplane/client.go
```diff
@@ -78,16 +89,82 @@ func (c *client) Authorize(req *api.AuthorizationRequest) (*remoteServerAuthoriz
...
+
+// heartbeatMonitor checks all peers for responsiveness, every fixed amount of time.
+func (c *client) heartbeatMonitor() {
+	c.logger.Info("Starting sending heartbeats to peer")
+	ticker := time.NewTicker(heartbeatInterval)
+	defer ticker.Stop()
+	// get heartbeat from all the peers
+	for {
+		t := time.Now()
+		if c.active || (!c.active && (t.Sub(c.lastSeen) > heartbeatRetransmissionTime)) {
```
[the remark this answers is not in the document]
kfirtoledo: Duration to the wait not for the time

pkg/controlplane/client.go
```diff
@@ -78,16 +89,82 @@ func (c *client) Authorize(req *api.AuthorizationRequest) (*remoteServerAuthoriz
 	return resp, nil
 }
 
+// Active returns if the peer is active or not.
+func (c *client) Active() bool {
```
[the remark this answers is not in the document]
kfirtoledo: Done

pkg/controlplane/client.go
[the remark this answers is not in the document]
kfirtoledo: Why this is a private function that can call the private variables of the class

pkg/controlplane/client.go
```diff
@@ -78,16 +89,82 @@ func (c *client) Authorize(req *api.AuthorizationRequest) (*remoteServerAuthoriz
...
 // newClient returns a new Peer API client.
 func newClient(peer *store.Peer, tlsConfig *tls.Config) *client {
 	clients := make([]*jsonapi.Client, len(peer.Gateways))
 	for i, endpoint := range peer.Gateways {
 		clients[i] = jsonapi.NewClient(endpoint.Host, endpoint.Port, tlsConfig)
 	}
-	return &client{
+	c := &client{
 		clients: clients,
+		active:  true,
```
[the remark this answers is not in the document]
kfirtoledo: You are right, but I don't want to create a special code for the first time we start the heartbeat and I think it is a reasonable assumption that if you add a peer it is already active, and if not, the GW will update the status after the first interval.

pkg/controlplane/client.go
```diff
@@ -78,16 +89,86 @@ func (c *client) Authorize(req *api.AuthorizationRequest) (*remoteServerAuthoriz
...
+	// copy peer clients array aside
+	peerClients := make([]*jsonapi.Client, len(c.clients))
+	c.lock.RLock()
+	copy(peerClients, c.clients)
+	c.lock.RUnlock()
+
+	for _, client := range peerClients {
+		serverResp, err := client.Get(api.HeartbeatPath)
+		if err != nil {
+			retErr = err
```
elevran, line 121: In the case of wanting to test multiple gateways of a peer, [`errors.Join()`](https://pkg.go.dev/errors@master#Join) may be useful way to collect them.
This applies to adding above `err` as well as `fmt.Errorf()` below

pkg/controlplane/client.go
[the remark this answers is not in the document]
elevran: you can create a code block to scope the locking to the copy only

kfirtoledo:
@elevran, PTAL

pkg/controlplane/client.go
```diff
@@ -78,16 +90,90 @@ func (c *client) Authorize(req *api.AuthorizationRequest) (*remoteServerAuthoriz
...
+func (c *client) IsActive() bool {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	return c.active
+}
+
+// setActive the peer status (active or not).
+func (c *client) setActive(active bool) {
+	c.lock.Lock()
+	defer c.lock.Lock()
```
elevran, line 103:
```suggestion
	defer c.lock.Unlock()
```

pkg/controlplane/client.go
```diff
@@ -78,16 +90,90 @@ func (c *client) Authorize(req *api.AuthorizationRequest) (*remoteServerAuthoriz
...
+	{
+		c.lock.RLock()
+		defer c.lock.RUnlock()
+		copy(peerClients, c.clients)
+	}
+
+	for _, client := range peerClients {
+		serverResp, err := client.Get(api.HeartbeatPath)
+		if err != nil {
+			retErr = errors.Join(retErr, err)
```
elevran, line 124: you can use `continue` here and then drop/unindent the else portion

pkg/controlplane/client.go
```diff
@@ -78,16 +90,90 @@ func (c *client) Authorize(req *api.AuthorizationRequest) (*remoteServerAuthoriz
...
+
+	return retErr // Return an error if all client targets are unreachable
+}
+
+// heartbeatMonitor checks all peers for responsiveness, every fixed amount of time.
+func (c *client) heartbeatMonitor() {
+	c.logger.Info("Starting sending heartbeats to peer")
+	ticker := time.NewTicker(heartbeatInterval)
+	defer ticker.Stop()
+	// get heartbeat from all the peers
```
elevran, line 142:
```suggestion
```

pkg/controlplane/client.go
```diff
@@ -78,16 +90,90 @@ func (c *client) Authorize(req *api.AuthorizationRequest) (*remoteServerAuthoriz
...
+				serverResp.Status, serverResp.Body))
+		}
+	}
+
+	return retErr // Return an error if all client targets are unreachable
+}
+
+// heartbeatMonitor checks all peers for responsiveness, every fixed amount of time.
+func (c *client) heartbeatMonitor() {
+	c.logger.Info("Starting sending heartbeats to peer")
```
elevran, line 139:
```suggestion
	c.logger.Info("Start sending heartbeat requests to peer")
```

elevran, line 139: might be worthwhile to keep the peer name on the `client` so it can be used for more informative logging message

pkg/controlplane/client.go
```diff
@@ -78,16 +90,90 @@ func (c *client) Authorize(req *api.AuthorizationRequest) (*remoteServerAuthoriz
...
+	c := &client{
+		clients:  clients,
+		active:   false,
+		lastSeen: time.Time{},
 		logger: logrus.WithFields(logrus.Fields{
 			"component": "peer-client",
 			"peer":      peer}),
 	}
+
+	go c.heartbeatMonitor()
```
elevran, line 177: this go routine does not terminate, correct? What if the peer is removed?

pkg/controlplane/server/http/heartbeat.go
```diff
@@ -0,0 +1,19 @@
...
+
+func (s *Server) addHeartbeatHandler() {
+	r := s.Router()
+
+	r.Get(api.HeartbeatPath, s.Heartbeat)
+}
+
+// Heartbeat returns a response for heartbeat checks from remote peers.
+func (s *Server) Heartbeat(w http.ResponseWriter, _ *http.Request) {
+	// Response
```
elevran, line 17:
```suggestion
```

elevran (changes requested):

pkg/controlplane/client.go
kfirtoledo: So you remove the client and the go routine is part of the client so it will be terminated.

pkg/controlplane/client.go
kfirtoledo: It is part of the logger see line 175

kfirtoledo:
@elevran PTAL

kfirtoledo:
@elevran PTAL

elevran (approved):

PR #55, opened 2023-10-02
tests/iperf3 : support new control plane for k8s tests

kfirtoledo:
Allow to run the iperf3 test and e2e connectivity test with the new controlplane.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,15 +85,17 @@ clean-tests:
 #------------------------------------------------------
 # Run Targets
 #------------------------------------------------------
+CP ?= old
+
 unit-tests:
 	@echo "Running unit tests..."
 	$(GO) test -v -count=1 ./pkg/...  -json -cover | tparse --all
 
-tests-e2e: clean-tests 	docker-build 
-	$(GO) test -p 1 -timeout 30m -v -tags e2e ./tests/e2e/connectivity/...
+tests-e2e: build docker-build
+	$(GO) test -p 1 -timeout 30m -v -tags e2e ./tests/e2e/connectivity/... --controlplane=$(CP)
 
-tests-iperf3: clean-tests 	docker-build 
-	$(GO) test -p 1 -timeout 30m -v -tags e2e ./tests/e2e/iperf3/...
+tests-iperf3: build docker-build 
+	$(GO) test -p 1 -timeout 30m -v -tags e2e ./tests/e2e/iperf3/... --controlplane=$(CP)
 
 run-gwctl:
 	@./bin/gwctl
--- a/Makefile
+++ b/Makefile
@@ -85,17 +85,15 @@ clean-tests:
 #------------------------------------------------------
 # Run Targets
 #------------------------------------------------------
-CP ?= old
-
 unit-tests:
 	@echo "Running unit tests..."
 	$(GO) test -v -count=1 ./pkg/...  -json -cover | tparse --all
 
 tests-e2e: build docker-build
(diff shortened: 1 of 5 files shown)


tests/e2e/connectivity/basic_test.go
```diff
@@ -30,9 +31,11 @@ var (
 	gwctl2             *client.Client
 )
 
+var cpType = flag.String("controlplane", "old", "Check which control-plane to use")
```
praveingk, line 34: Do we need this? we do not plan to keep the old control plane or make any changes on it. Why not keep the tests only the new control plane only.

tests/e2e/utils/kind.go
```diff
@@ -155,58 +150,101 @@ func DeleteCluster(name string) error {
...
+			return err
+		}
 
-	dpPod, _ := GetPodNameIP("dataplane")
-	cPortStr := strconv.Itoa(int(cPort))
-	err = runCmd("kubectl create service nodeport dataplane --tcp=" + cPortLocal + ":" + cPortLocal + " --node-port=" + cPortStr)
-	if err != nil {
-		return err
-	}
+		err = runCmd("kubectl apply -f " + manifests + "mbg-role.yaml")
```
praveingk, line 192: Can we rename the file mbg-role to clusterlink-role?

Makefile
```diff
@@ -83,15 +83,16 @@ clean-tests:
...
 	$(GO) test -v -count=1 ./pkg/...  -json -cover | tparse --all
 
-tests-e2e: clean-tests 	docker-build 
-	$(GO) test -p 1 -timeout 30m -v -tags e2e ./tests/e2e/connectivity/...
+tests-e2e: build docker-build
+	$(GO) test -p 1 -timeout 30m -v -tags e2e ./tests/e2e/connectivity/... --controlplane=$(CP)
 
-tests-iperf3: clean-tests 	docker-build 
-	$(GO) test -p 1 -timeout 30m -v -tags e2e ./tests/e2e/iperf3/...
+tests-iperf3: build docker-build
```
praveingk, line 94: Can we merge the tests? tests-e2e to run iperf tests too
Otherwise, we would be performing build redundantly if we are running both the tests.

praveingk (changes requested):

tests/e2e/connectivity/basic_test.go
kfirtoledo: yes I will remove it when we decide to remove the old controlplane

Makefile
kfirtoledo: So why not use only the iperf3 test?

tests/e2e/utils/kind.go
kfirtoledo: I will remove it should be part of the cl-adm

Makefile
praveingk: Using iperf3 for just testing connectivity and policy test seems like an overkill. I thought iperf3 was more for benchmarking performance, which we want to do part of each test. My problem is with having to build each time for running a test especially if both tests are going to be run.  Ideally, we can use the connectivity test to test policies, load balancing, etc in future.. iperf3 for only performance benchmarking.

elevran:
@kfirtoledo @praveingk if this PR is still ongoing please close on the open tasks (creating follow up PRs or issue as need).
Otherwise, if it is no longer needed/planned - please close

praveingk (approved):

PR #56, opened 2023-10-02
util/log: Add logutills to the new controlplane

kfirtoledo:
Move logutils to the util/log folder and remove the boolean flag from the SetLog function.
Use logutils.SetLog for setting the log in the cl-controlplane.


pkg/util/log/logutils.go
```diff
@@ -13,16 +13,16 @@ import (
...
 	logrusFieldlessStackJump = 6
 )
 
-func SetLog(logLevel string, logfile bool, logFileName string) {
-	usr, _ := user.Current()
-	logFileFullPath := path.Join(usr.HomeDir, projectFolder, logFileName)
-	CreateProjectfolder(logFileFullPath)
-	if logfile {
+func SetLog(logLevel string, logFileName string) {
+	if logFileName != "" {
```
elevran, line 21: is there a way to log to stdout and not a file?
It seems that if a path is not provided you use a default path under the home directory.

pkg/util/log/logutils.go
```diff
@@ -49,11 +50,11 @@ func SetLog(logLevel string, logfile bool, logFileName string) {
 	})
 }
 
-type MyFormatter struct {
+type myFormatter struct {
```
elevran, line 53: Please avoid types and variables with their name starting `My...`.
It is in the open so `My` seems inappropriate to me.
If there is a specific thing the Formatter does (e.g., colored output, ...), consider using that.

elevran (changes requested):

pkg/util/log/logutils.go
kfirtoledo: If the file name id empty it will print to the stdout

pkg/util/log/logutils.go
kfirtoledo: Done

elevran (approved):

pkg/util/log/logutils.go
orozery: Where do you see that?
If no log file is set I think it should output to stderr.
This is what logrus defaults to, so in that case you don't need to call `logrus.SetOutput` at all.

orozery (changes requested):
Two further comments:
1. I don't like the `getCurrentPosition` function. It looks hacky to me. I would at least make it run only if the logging level is set to debug.
2. Can we move the log utils to `pkg/util/log.go`?

kfirtoledo:
> Two further comments:
> 
> 1. I don't like the `getCurrentPosition` function. It looks hacky to me. I would at least make it run only if the logging level is set to debug.
> 2. Can we move the log utils to `pkg/util/log.go`?
@orozery, PTAL
1. The function is running just for error or fatal messages, I changed the name of the function.
2. Already was util/log change the filename to log.go instead of logutils.go

pkg/util/log/log.go
```diff
@@ -49,21 +50,22 @@ func SetLog(logLevel string, logfile bool, logFileName string) {
 	})
 }
 
-type MyFormatter struct {
+type myFormatter struct {
```
elevran, line 53: I thought this was previously changed to drop the `my`

pkg/util/log/log.go
```diff
@@ -80,7 +82,8 @@ start:
 	return function, file, lineNumber
 }
 
-func CreateProjectfolder(filePath string) {
+// createLogfolder create the log folder if not exists
+func createLogfolder(filePath string) {
```
elevran, line 86: - Why do you need to `Stat` the file?  Won't `os.MkDirAll` just do the right thing (ignores existing directories and creates all in path).

```suggestion
func createLogFolder(filePath string) {
```

elevran (changes requested):

pkg/util/log/log.go
kfirtoledo: Yes, my mistake.

kfirtoledo:
@elevran, @orozery  PTAL

elevran (approved):

cmd/cl-controlplane/app/server.go
```diff
@@ -58,27 +58,8 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
...
-		f, err := os.OpenFile(o.LogFile, os.O_APPEND|os.O_CREATE|os.O_RDWR, 0666)
-		if err != nil {
-			return fmt.Errorf("unable to open log file: %v", err)
-		}
-
-		defer func() {
-			if err := f.Close(); err != nil {
-				log.Errorf("Cannot close log file: %v", err)
-			}
-		}()
```
orozery, line 71: This cleanup is missing.
Maybe you can allow `SetLog` to accept an open file?

kfirtoledo: Done, 77951bbaa193166b9b0d755b51fb1f4785f42eb6

kfirtoledo:
@orozery PTAL

orozery (approved):
Also see linting errors.

PR #58, opened 2023-10-02
Custom Go-based dataplane for xDS control plane APIs

praveingk:
Re-creating based on https://github.com/clusterlink-net/clusterlink/pull/44 
@elevran : Addressed most of the comments from the previous PR, except the dataplane API refactoring (using controller) suggested by @orozery which would be done as another PR after removing the old dataplane code.


praveingk:
Created an issue to address the comments regarding using controller for dataplane:  https://github.com/clusterlink-net/clusterlink/issues/62

praveingk:
@elevran As per our discussion, I have removed the CLI flag for the cl-dataplane, and moved the go dataplane to cl-go-dataplane binary. cl-dataplane will launch envoy and cl-go-dataplane will launch go dataplane. PTAL

pkg/dataplane/client/xds.go
```diff
@@ -0,0 +1,76 @@
...
+		err = f.Run()
+		x.logger.Infof("Fetcher '%s' stopped: %v.", resourceType, err)
+	}
+}
+
+// Run starts the running xDS client which fetches clusters and listeners from the controlplane.
+func (x *XDSClient) Run() error {
+	var wg sync.WaitGroup
+	wg.Add(len(x.fetchers))
+	go func() {
```
elevran, line 53: suggestion:
consider using a for loop on elements of an array of `[resource.ClusterType, resource.ListenerType]`
and then you can fold the two go functions into a single code snippet.
If we add (e.g.,) `resource.SecretType` in the future, it becomes an easy addition with no code duplication

pkg/dataplane/server/dataplane.go
```diff
@@ -0,0 +1,77 @@
...
+
+// Dataplane implements the server and api client which sends authorization to the control plane
+type Dataplane struct {
+	ID                 string
+	peerName           string
+	router             *chi.Mux
+	apiClient          *http.Client
+	parsedCertData     *util.ParsedCertData
+	controlplaneTarget string
+	clusterMap         map[string]*cluster.Cluster
```
elevran, line 25: there is no need, IMO, to add the type name to variable names (`clusterMap`, `listenerChan`, etc).

pkg/dataplane/server/forwarder.go
```diff
@@ -0,0 +1,112 @@
...
+				return err
+			}
+			break
+		}
+	}
+	f.closeConnections()
+	return nil
+}
+
+func (f *forwarder) appToPeer() error {
```
elevran, line 51: consider using `workload` instead of `app`?

pkg/dataplane/server/forwarder.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+		_, err = f.appConn.Write(bufData[:numBytes]) // TODO: track actually written byte count
+		if err != nil {
+			if err != io.EOF { // don't log EOF
+				return err
+			}
+			break
+		}
+	}
+	f.closeConnections()
```
elevran, line 47: consider calling shutdown read/write instead of calling close on the connections. You may end up tearing down connections before all data has been transmitted (i.e., workload->gw closed, but still have some slower sending on gw->peer as it is over the WAN. CloseConnections will shutdown both)

pkg/dataplane/server/forwarder.go
```diff
@@ -0,0 +1,112 @@
...
+	wg.Add(1)
+	go func() {
+		wg.Done()
+		err := f.peerToApp()
+		if err != nil {
+			f.logger.Errorf("End of peer to app connection %v.", err)
+		}
+	}()
+
+	wg.Wait()
```
elevran, line 104: I think you should only close connections when `wg.Wait()` is done and not on every error. See previous note on using CloseRead/CloseWrite for signalling done with data

pkg/dataplane/server/forwarder.go
elevran: Go has a [tcpConn.CloseRead](https://pkg.go.dev/net#TCPConn.Close)

pkg/dataplane/server/listener.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		d.logger.Infof("Error Listen to port %v", err)
+		return
+	}
+	go func() {
+		if err = d.serveEgressConnections(name, acceptor); err != nil {
+			d.logger.Errorf("Failed to serve egress connection on %s: %+v.", listenTarget, err)
+		}
+	}()
+	<-d.listenerChan[name]
```
elevran, line 34: what is the channel synchronizing?

elevran (changes requested):

cmd/cl-dataplane/app/server.go
```diff
@@ -29,6 +29,8 @@ const (
 type Options struct {
 	// ControlplaneHost is the IP/hostname of the controlplane.
 	ControlplaneHost string
+	// Type is the dataplane type.
+	Type string
```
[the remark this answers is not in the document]
praveingk: Yes, it was stray variable unused. Removed it now.

cmd/cl-go-dataplane/app/server.go
```diff
@@ -0,0 +1,161 @@
...
+	// Start xDS client, if it fails to start we keep retrying to connect to the controlplane host
+	tlsConfig := parsedCertData.ClientConfig(cpapi.GRPCServerName(peerName))
+	xdsClient := dpclient.NewXDSClient(dataplane, controlplaneTarget, tlsConfig)
+	err := xdsClient.Run()
+	return fmt.Errorf("xDS Client stopped : %v", err)
+}
+
+// Run the dataplane.
+func (o *Options) Run() error {
+	// set log file
```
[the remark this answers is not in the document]
praveingk: Sure. Will do.

cmd/cl-go-dataplane/app/server.go
```diff
@@ -0,0 +1,161 @@
...
+		return fmt.Errorf("expected peer certificate to contain a single DNS name, but got %d", len(dnsNames))
+	}
+
+	peerName, err := api.StripServerPrefix(dnsNames[0])
+	if err != nil {
+		return err
+	}
+
+	// generate random dataplane ID
+	dataplaneID := uuid.New().String()
```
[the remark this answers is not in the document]
praveingk: Agree Etai. I think this needs a uniform change in envoy dataplane also. Hence, will create an issue for this to be addressed in future.

pkg/dataplane/client/fetcher.go
```diff
@@ -0,0 +1,102 @@
...
+		if err != nil {
+			return err
+		}
+		f.logger.Debugf("Listener : %s.", l.Name)
+		listenerName := strings.TrimPrefix(l.Name, api.ImportListenerPrefix)
+		err = f.dataplane.AddListener(listenerName, l)
+		if err != nil {
+			continue
+		}
+		go func() {
```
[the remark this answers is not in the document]
praveingk: Agree 👍🏽

pkg/dataplane/client/xds.go
```diff
@@ -0,0 +1,76 @@
...
+	for {
+		time.Sleep(5 * time.Second)
+		conn, err := grpc.Dial(x.controlplaneTarget, grpc.WithTransportCredentials(credentials.NewTLS(x.tlsConfig)))
+		if err != nil {
+			x.logger.Errorf("Failed to dial controlplane xDS server: %v.", err)
+			continue
+		}
+		x.logger.Infof("Successfully connected to the controlplane xDS server.")
+
+		f, err := newFetcher(context.Background(), conn, resourceType, x.dataplane)
```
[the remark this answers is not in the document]
praveingk: @orozery suggested to use separate graph connections for both fetchers. It tries to create a connection again only upon failure of the fetcher. Otherwise, at a certain point there are two active connections only.

pkg/dataplane/server/dataplane.go
praveingk: Agree 👍🏽

cmd/cl-go-dataplane/app/server.go
```diff
@@ -0,0 +1,157 @@
...
+
+// RequiredFlags are the names of flags that must be explicitly specified.
+func (o *Options) RequiredFlags() []string {
+	return []string{"controlplane-host"}
+}
+
+// Run the go dataplane.
+func (o *Options) runGoDataplane(peerName, dataplaneID string, parsedCertData *util.ParsedCertData) error {
+
+	controlplaneTarget := o.ControlplaneHost + ":" + strconv.Itoa(cpapi.ListenPort)
```
elevran, line 63: Prefer use of `net.JoinHostPort`?

pkg/dataplane/server/listener.go
praveingk: It runs and waits until set by deleteListener

pkg/dataplane/server/forwarder.go
praveingk: I tried moving the closeConnection after wg.wait. The side-effect is that the one way connection doesn't tear down automatically, and for some reason iperf3 doesn't seem to work with that behavior. Hence, I have tried adding a close signal, which will exit the loop in a graceful way and finally close connection. Hence, to exit the loop, we need to add a read deadline so that it doesn't wait on the connection.

pkg/dataplane/server/forwarder.go
praveingk: I have added the change in the latest commit https://github.com/clusterlink-net/clusterlink/pull/58/commits/67b3ae3fc935f8cee2282616eb04bbbc91bbd759

pkg/dataplane/server/forwarder.go
elevran: here's the pseudocode of the logic I had in mind (simplified):
```go
   nb, err := c1.Read(...)
   if err != nil { // done reading from c1
      if nb > 0 { // write last remaining bytes in c1->c2, if any
         c2.Write(...)
      }
      // signal c2 that we're done writing - this will send a FIN and allow upstream to close c2 
      c2.CloseWrite()
      break // done with the read/write loop, signal the wg outside so that connections can be closed when it's done
   }
   // continue normal processing for read/write when no error on read
   nb, err = c2.Write(...) 
   ...
```

praveingk: I did a quick try to use TCPConn. However, Hijacker connection is unable to be cast to TCPConn. Hence it failed to establish connection.

elevran: Since we started with an HTTP socket, built on top of TCP, the hijacked `net.Conn` connection should be a `net.TCPConn`. I don't see how it would fail the cast/type-check. For debug, can you inspect the type of connection object returned?

elevran (approved):
We can address the connection tear down with a separate PR (if it is indeed feasible).
For now - code is approved. Please ensure it passes all checks (I will confirm the DCO)

pkg/dataplane/server/listener.go
```diff
@@ -0,0 +1,89 @@
...
+			d.logger.Error("Failed to accept egress connection", err)
+			return err
+		}
+
+		d.logger.Debugf("Received an egress connection at listener for imported service %s from %s.", name, conn.RemoteAddr().String())
+		d.logger.Debugf("Connection: %+v.", conn)
+
+		targetPeer, accessToken, err := d.getEgressAuth(name, strings.Split(conn.RemoteAddr().String(), ":")[0])
+		if err != nil {
+			d.logger.Error("Failed egress authorization", err)
```
orozery, line 52: IMO If the error is due to a policy, this should be recorded at info level, not error level.

pkg/dataplane/server/listener.go
```diff
@@ -0,0 +1,89 @@
...
+			conn.Close()
+			continue
+		}
+		d.logger.Infof("Received auth from controlplane: target peer: %s with %s", targetPeer, accessToken)
+		tlsConfig := d.parsedCertData.ClientConfig(dpapi.DataplaneServerName(strings.TrimPrefix(targetPeer, api.RemotePeerClusterPrefix)))
```
orozery, line 57: I think it's neater to get the server name from `d.clusters[targetPeer]` instead of trimming the `targetPeer` string.

pkg/dataplane/server/listener.go
```diff
@@ -0,0 +1,89 @@
...
+
+		go func() {
+			err := d.initiateEgressConnection(targetPeer, accessToken, conn, tlsConfig)
+			if err != nil {
+				d.logger.Error("Failed to initiate egress connection ", err)
```
orozery, line 62:
```suggestion
				d.logger.Errorf("Failed to initiate egress connection:  %v.", err)
```

pkg/dataplane/server/listener.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return "", "", err
+	}
+	egressAuthReq.Close = true
+
+	egressAuthReq.Header.Add(api.ClientIPHeader, sourceIP)
+	egressAuthReq.Header.Add(api.ImportHeader, name)
+	egressAuthResp, err := d.apiClient.Do(egressAuthReq)
+	if err != nil {
+		d.logger.Error("Unable to send auth/egress request ", err)
```
orozery, line 80:
```suggestion
		d.logger.Errorf("Unable to send auth/egress request: %v.", err)
```

Actually, I think this line could be removed, since the caller already logs the error.

pkg/dataplane/server/listener.go
```diff
@@ -0,0 +1,89 @@
...
+		return "", "", err
+	}
+	defer egressAuthResp.Body.Close()
+	if egressAuthResp.StatusCode != http.StatusOK {
+		d.logger.Infof("Failed to obtained egress authorization: %s", egressAuthResp.Status)
```
orozery, line 85:
```suggestion
		d.logger.Infof("Failed to obtain egress authorization: %s", egressAuthResp.Status)
```

Also, can be removed.

orozery (changes requested):

pkg/dataplane/server/listener.go
elevran: [Go fmt.Errorf docs](https://pkg.go.dev/fmt#Errorf) suggest using `%w` for errors. Otherwise, `%s` and `%v` are both interpreting the argument as a String.

pkg/dataplane/server/server.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		d.logger.Errorf("Unable to get cluster target: %v.", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	// hijack connection
+	peerConn, err := d.hijackConn(w)
+	if err != nil {
+		d.logger.Error("hijacking failed ", err)
```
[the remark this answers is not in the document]
elevran: I think errors should not contain punctuation marks as they may be wrapped and concatenated in the call stack.

pkg/dataplane/server/server.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/clusterlink-net/clusterlink/pkg/utils/netutils"
+)
+
+const (
+	httpSchemaPrefix = "https://"
+)
+
+// StartDataplaneServer starts the Dataplane server
+func (d *Dataplane) StartDataplaneServer(dataplaneServerAddress string) error {
+	d.logger.Infof("Dataplane server starting  at %s", dataplaneServerAddress)
```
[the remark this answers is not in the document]
elevran: @orozery @praveingk 
errors should 
1. not contain punctuation marks
2. start with a capital letter

The [language code review guidelines](https://github.com/golang/go/wiki/CodeReviewComments#error-strings), reasons that errors may be wrapped and/or concatenated as they're passed up the stack

pkg/dataplane/server/server.go
```diff
@@ -0,0 +1,179 @@
...
+	err := sniProxy.Listen(dataplaneListenAddress)
+	if err != nil {
+		return fmt.Errorf("unable to create listener for server on %s: %v",
+			dataplaneListenAddress, err)
+	}
+	return sniProxy.Serve()
+}
+
+func (d *Dataplane) addAuthzHandlers() {
+	d.router.Post(cpapi.DataplaneIngressAuthorizationPath, d.dataplaneIngressAuthorize)
```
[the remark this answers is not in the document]
praveingk: Oh I see, I thought that change you had suggested was related to the refactoring. Let me change to "/" and verify, if it works and change.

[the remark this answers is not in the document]
praveingk: Verified and changed it to "/" instead of the specific path.

pkg/dataplane/server/listener.go
praveingk: The server name of the remote peer isn't available in the cluster struct. It has only the following :
```
		endpoints:{
			lb_endpoints:{
				endpoint:{
					address:{
						socket_address:{
							address:\"peer1-dataplane\"
							port_value:443
						}
					}
					hostname:\"dataplane.peer1:443\"
				}
			}
		}
```
In this case, we need the target peer string as "peer1" which is named as "remote-peer-peer1"

pkg/dataplane/server/listener.go
praveingk: Yes, you are right. Changed to info.

pkg/dataplane/client/xds.go
```diff
@@ -0,0 +1,73 @@
...
+	dataplane          *server.Dataplane
+	controlplaneTarget string
+	tlsConfig          *tls.Config
+	errors             []error
+	logger             *logrus.Entry
+}
+
+func (x *XDSClient) runFetcher(resourceType string) error {
+	for {
+		time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
praveingk: The problem is when we do continue, sleep doesn't execute if it's in the end of the loop. In those cases, we might need to  execute sleep before continue, as well as in the end of the loop. That's fine with me, if its cleaner.

pkg/dataplane/server/server.go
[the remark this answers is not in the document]
praveingk: > Verified and changed it to "/" instead of the specific path.

Actually, my verification was incorrect earlier. Upon revalidating. This change doesn't work. It needed the following commit as well : https://github.com/clusterlink-net/clusterlink/pull/58/commits/7b363c0cc9e6e69f38a59ec4ebf88a7370698f8a

pkg/dataplane/server/forwarder.go
```diff
@@ -0,0 +1,134 @@
...
+
+const (
+	dataBufferSize = 64 * 1024
+	readDeadline   = 500 * time.Millisecond
+)
+
+type forwarder struct {
+	workloadConn net.Conn
+	peerConn     net.Conn
+	closeSignal  bool
```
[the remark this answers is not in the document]
praveingk: This is done.

praveingk:
@orozery  I have addressed the comments and verified the working.

praveingk:
> We can address the connection tear down with a separate PR (if it is indeed feasible). For now - code is approved. Please ensure it passes all checks (I will confirm the DCO)

I have created https://github.com/clusterlink-net/clusterlink/issues/73 to track this.

pkg/dataplane/server/server.go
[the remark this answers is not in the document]
orozery: @elevran The opposite holds for logging messages.

pkg/dataplane/client/xds.go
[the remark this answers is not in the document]
orozery: How about `for ;; time.Sleep(5 * Second)` ?
Another option is to move the loop body to another function (could be defined inside the current one).

pkg/dataplane/server/listener.go
orozery: You actually need `dataplane.peer1:443` which is available above.

pkg/dataplane/server/server.go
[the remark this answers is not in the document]
elevran: 👍 
So logging messages follow full sentence structure (start capitalized and end in punctuation) and error messages should not (i.e. start lower case and not end with punctuation).
Created #76 to track this.

pkg/dataplane/server/listener.go
praveingk: ok, I can trim the server name from hostname 👍🏽

praveingk: https://github.com/clusterlink-net/clusterlink/pull/58/commits/a1510e4156f08f733af8fbf64502684b3a0ba34e

cmd/cl-adm/cmd/create/create_peer.go
```diff
@@ -23,12 +23,15 @@ type PeerOptions struct {
...
 	Dataplanes uint16
+	// DataplaneType is the type of dataplane to create (envoy or go-based)
+	DataplaneType string
 }
 
 // AddFlags adds flags to fs and binds them to options.
 func (o *PeerOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.Name, "name", "", "Peer name.")
 	fs.Uint16Var(&o.Dataplanes, "dataplanes", 1, "Number of dataplanes.")
+	fs.StringVar(&o.DataplaneType, "dataplaneType", "envoy", "Type of dataplane")
```
orozery, line 34: Few suggestions:
- Rename CLI arg from `dataplaneType` to `dataplane-type`.
- Add possible values to description (i.e. "Type of dataplane. Supported values: "envoy" (default), "go".
- Define the possible values as const strings. e.g:
```
	// DataplaneTypeEnvoy represents an envoy-type dataplane.
	DataplaneTypeEnvoy = "envoy"
	// DataplaneTypeGo represents a go-type dataplane.
	DataplaneTypeGo = "go"
```

cmd/cl-adm/templates/docker.go
```diff
@@ -38,6 +39,19 @@ cl-dataplane \
...
 --log-file {{.persistencyDirectoryMountPath}}/log.log
+{{ else }}
+docker run -itd \
+--name {{.peer}}-dataplane \
+-v $FABRIC_DIR/{{.fabricCAPath}}:{{.dataplaneCAMountPath}} \
+-v $FABRIC_DIR/{{.dataplaneCertPath}}:{{.dataplaneCertMountPath}} \
+-v $FABRIC_DIR/{{.dataplaneKeyPath}}:{{.dataplaneKeyMountPath}} \
+-v $FABRIC_DIR/{{.dataplanePersistencyDirectory}}:{{.persistencyDirectoryMountPath}} \
+cl-go-dataplane \
+cl-go-dataplane \
```
orozery, line 50: Can you reduce the if-else just to these 2 lines?

orozery (changes requested):
Can you also change the github actions workflow to run `tests/k8s.sh` with the go dataplane?

praveingk:
Done in https://github.com/clusterlink-net/clusterlink/pull/58/commits/48f9ac121cd005f0ef282f83aa219ae2b3372106

tests/k8s.sh
```diff
@@ -4,6 +4,7 @@ set -ex
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 TEST_DIR=$(mktemp -d)
 CLADM=$SCRIPT_DIR/../bin/cl-adm
+TYPE="${1:-envoy}"
```
orozery, line 7: Can you rename this to `DATAPLANE_TYPE`?

cmd/cl-adm/templates/config.go
```diff
@@ -63,9 +73,16 @@ func (c Config) TemplateArgs() (map[string]interface{}, error) {
...
+	if c.DataplaneType == DataplaneTypeGo {
+		dataplaneTypeEnvoy = false
+	}
+
 	return map[string]interface{}{
-		"peer":       c.Peer,
-		"dataplanes": c.Dataplanes,
+		"peer":               c.Peer,
+		"dataplanes":         c.Dataplanes,
+		"dataplaneTypeEnvoy": dataplaneTypeEnvoy,
```
orozery, line 85: See my suggestion in Slack on passing `dataplaneType` in addition to all `dataplaneType*` strings.

orozery (changes requested):

orozery (commented):
@praveingk Can you also add a switch case check to `PeerOptions::Run()` to verify that `DataplaneType` has a valid value?

praveingk:
> @praveingk Can you also add a switch case check to `PeerOptions::Run()` to verify that `DataplaneType` has a valid value?

Its done here : https://github.com/clusterlink-net/clusterlink/pull/58/commits/0faaeb21fc59210e1a9f40ec9200b5f7618cf40f

elevran:
@orozery reading through the conversation it seems all requests have been addressed. Is there anything else that needs to be done? If not - please mark approval so it can be merged

orozery (approved):

PR #59, opened 2023-10-03
cmd/gwctl: Add get all command

kfirtoledo:
Add get all commands that prints all: peers, imports, exports, bindings and policies.


cmd/gwctl/subcommand/gwctl.go
```diff
@@ -114,3 +115,54 @@ func (o *stateGetOptions) run() error {
...
+
+type allGetOptions struct {
+	myID string
+}
+
+// AllGetCmd - get all gateways objects.
+func AllGetCmd() *cobra.Command {
+	o := allGetOptions{}
+	cmd := &cobra.Command{
+		Use:   "all",
```
elevran, line 127: `gwctl get config-dump` ?

elevran (approved):

cmd/gwctl/subcommand/gwctl.go
kfirtoledo: I thought to do it similar to kubectl get all

PR #60, opened 2023-10-03
cicd: Enable additional linters

elevran:
This PR enables additional linters. The list represents (what I think is) a reasonable set of linters that should be run in CICD.
However, some linters are currently disabled in the `golangci.yaml` configuration as the existing code does not pass the checks.

Future PRs can update the linter set or their configuration by removing the commented out lines and setting `issues.new: false` in `golangci.yaml`  and then running `make lint` and fixing any reported issues. This work will tracked in a #61.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -16,8 +16,7 @@ linters:
   # set to true to run only fast linters (e.g., for pre-commit)
   fast: false
 
-  disable-all: false
-
+  disable-all: true
   enable:
     - asasalint
     - asciicheck
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -45,15 +45,15 @@ linters:
     - importas
     - ineffassign
     - makezero
-    # - misspell
+    - misspell
     - nakedret
     # - nestif
     - nilerr
     # - nilnil
     # - noctx
     # - nolintlint
-    # - nonamedreturns
-    # - nosprintfhostport
+    - nonamedreturns
+    - nosprintfhostport
     - prealloc
     - predeclared
     - promlinter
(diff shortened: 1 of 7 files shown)


praveingk (approved):

Makefile
```diff
@@ -32,16 +32,16 @@ format: fmt
 fmt: format-go tidy-go vet-go
 vet: vet-go
 
-lint:  ; $(info $(M) running linters...)
-	@golangci-lint run
+lint:  ; $(info running linters...)
+	@golangci-lint run --config=./.golangci.yaml ./...
```
kfirtoledo, line 36: I think this is not necessary.

.golangci.yaml
```diff
@@ -1,29 +1,140 @@
...
+    - asasalint
+    - asciicheck
+    - bidichk
     - bodyclose
-    - errcheck
+    - contextcheck
+    - durationcheck
+    - errcheck 
+    - errname
+    # - errorlint
```
kfirtoledo, line 29: I think you should  open a PR for all the linters you want to add in the future but wasn't able to do it now

Makefile
elevran: It is not necessary as this is the default. I don't think making it explicit hurts

.golangci.yaml
elevran: A PR would involve making all of the code changes to make each linter pass. Did you mean an issue? I opened #61 to track this wholesale, with needed linters already in this PR's config file only commented out.
Do you think anything else is needed?

kfirtoledo: That's ok

kfirtoledo (approved):

.golangci.yaml
```diff
@@ -1,29 +1,140 @@
+# Based on recommendations in https://olegk.dev/go-linters-configuration-the-right-version
+
 run:
+  # linter execution
   timeout: 30m
+  issues-exit-code: 1
+  concurrency: 4
+  # check tests as well
+  tests: true
```
orozery, line 9: I don't think this is desirable (and especially in the current case where e2e tests are defined as Go tests).
When you want to run linter, you just want to run linter.
Tests is a different target.

orozery (changes requested):

.golangci.yaml
elevran: Perhaps there's some confusion.
Setting `tests: true` runs Go **linters** on the test files, and does **not** run the tests themselves.
Since it's Go code, I see no reason not to lint them.

orozery: My bad :)

PR #71, opened 2023-10-08
tests/k8s.sh: Fix iperf server setup

orozery:
This PR fixes the iperf server setup.


elevran (approved):
question: is there a reason to run with `-t` when not running in automation?

elevran:
:man_shrugging:
Was just wondering if there was a reason to include it originally...

PR #72, opened 2023-10-09
Deployment/k8s: Add K8s deployment to the new controlplane

kfirtoledo:
This PR adds the k8s deployment to the new controlplane.
This includes support in:
1) services 
2) endpoints
3) translation of IP to labels


pkg/deployment/k8s/deployment.go
```diff
@@ -0,0 +1,227 @@
...
+		}
+	}
+
+	return nil, fmt.Errorf("pod with IP %s not found", podIP)
+}
+
+// NewDeployment returns a new Kubernetes deployment.
+func NewDeployment() (*Deployment, error) {
+	logger := logrus.WithField("component", "k8s-deployment")
+	cl, err := client.New(config.GetConfigOrDie(), client.Options{})
```
orozery, line 199: I don't think it's a good idea to use `OrDie`.

pkg/deployment/k8s/deployment.go
```diff
@@ -0,0 +1,227 @@
...
+	if err != nil {
+		return &Deployment{}, err
```
orozery, line 201:
```suggestion
		return nil, err
```

pkg/deployment/k8s/deployment.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	// Get namespace
+	var podList corev1.PodList
+	labelSelector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{
+		MatchLabels: map[string]string{"app": "cl-controlplane"}})
+	if err != nil {
+		return &Deployment{}, err
```
orozery, line 209:
```suggestion
		return nil, err
```

pkg/deployment/k8s/deployment.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	listOptions := &client.ListOptions{LabelSelector: labelSelector}
+	err = cl.List(context.Background(), &podList, listOptions)
+	if err != nil {
+		return &Deployment{}, err
+	}
+
+	if len(podList.Items) == 0 {
+		return &Deployment{}, fmt.Errorf("pod not found")
```
orozery, line 219:
```suggestion
		return nil, fmt.Errorf("pod not found")
```

pkg/deployment/k8s/deployment.go
```diff
@@ -0,0 +1,227 @@
...
+
+// NewDeployment returns a new Kubernetes deployment.
+func NewDeployment() (*Deployment, error) {
+	logger := logrus.WithField("component", "k8s-deployment")
+	cl, err := client.New(config.GetConfigOrDie(), client.Options{})
+	if err != nil {
+		return &Deployment{}, err
+	}
+
+	// Get namespace
```
orozery, line 204: I think this solution is neater:
https://stackoverflow.com/a/57809472

orozery (changes requested):
Regarding naming, I agree `Deployment` is confusing.
I think @praveingk suggested `RuntimeEnv`.
I personally prefer `Platform`.

cmd/cl-adm/templates/k8s.go
```diff
@@ -91,7 +91,7 @@ spec:
         - name: cl-controlplane
           image: cl-controlplane
           imagePullPolicy: IfNotPresent
-          args: ["--log-level", "info"]
+          args: ["--log-level", "info", "--deployment", "k8s"]
```
praveingk, line 94: deployment could be changed to runtimeEnv or platform as suggested by @orozery

pkg/deployment/k8s/deployment.go
```diff
@@ -0,0 +1,227 @@
...
+	d.logger.Infof("Deleting K8s endPoint %s", name)
+	return nil
+}
+
+// GetPodLabelsByIP returns all the labels that match the pod IP.
+func (d *Deployment) GetPodLabelsByIP(podIP string) (map[string]string, error) {
+	var podList corev1.PodList
+	err := d.client.List(context.Background(), &podList) // Adjust the label selector as needed
+	if err != nil {
+		fmt.Printf("GetPodLabelsByIP %s", err.Error())
```
praveingk, line 182: change to use d.logger

pkg/deployment/k8s/deployment.go
```diff
@@ -0,0 +1,227 @@
...
+		d.logger.Errorf("error occurred while deleting K8s endpoint %s:", err)
+		return err
+	}
+
+	d.logger.Infof("Deleting K8s endPoint %s", name)
+	return nil
+}
+
+// GetPodLabelsByIP returns all the labels that match the pod IP.
+func (d *Deployment) GetPodLabelsByIP(podIP string) (map[string]string, error) {
```
praveingk, line 178: In the earlier version of control plane, we had used kube informers.
I am unsure of the latency of this call since we have to get the entire pod list and iterate through it upon every incoming connection, which is very inefficient. Can you try using informers which are synced and kept in memory (https://github.com/clusterlink-net/clusterlink/blob/main/pkg/k8s/kubernetes/kubeinformer.go) . I am unaware of any downsides to this approach.

praveingk (changes requested):

pkg/deployment/k8s/deployment.go
kfirtoledo: I agree, but the main point here of this change is to use more high-level abstraction I will check if we can cache the pod lists also here. @elevran any suggestions.

pkg/deployment/k8s/deployment.go
```diff
@@ -0,0 +1,227 @@
...
+		},
+	}
+
+	err := d.client.Create(context.Background(), serviceSpec)
+	if err != nil {
+		d.logger.Errorf("error occurred while creating K8s service %s:", err)
+		return err
+	}
+
+	d.logger.Infof("Creating K8s service at %s:%d ", name, port)
```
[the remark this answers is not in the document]
kfirtoledo: I prefer without the space like every IP address and I print it only if it is succeed.

pkg/deployment/k8s/deployment.go
```diff
@@ -0,0 +1,227 @@
...
+				},
+			},
+			Type:     corev1.ServiceTypeClusterIP,
+			Selector: map[string]string{"app": dataplaneAppName},
+		},
+	}
+
+	err := d.client.Create(context.Background(), serviceSpec)
+	if err != nil {
+		d.logger.Errorf("error occurred while creating K8s service %s:", err)
```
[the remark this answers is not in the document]
kfirtoledo: Personally, I prefer to print in the place where the error has happened it is easier to debug

pkg/controlplane/instance.go
```diff
@@ -132,6 +132,22 @@ func (cp *Instance) CreateExport(export *cpstore.Export) error {
...
 		}
+		// Creates k8s endpoint and service for external service.
+		exSvc := export.ExportSpec.ExternalService
+		if exSvc.Host != "" && exSvc.Port != 0 {
+			err := cp.deployment.CreateEndpoint(export.Name, exSvc.Host, exSvc.Port)
+			if err != nil {
+				return err
+			}
+
+			err = cp.deployment.CreateService(export.Name, exSvc.Port, exSvc.Port)
```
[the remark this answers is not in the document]
kfirtoledo: The service point to the endpoint (they have the same name) and the endpoint point to the external service.

pkg/deployment/k8s/deployment.go
```diff
@@ -0,0 +1,227 @@
...
+	d.logger.Infof("Creating K8s service at %s:%d ", name, port)
+	return nil
+}
+
+// DeleteService deletes a service.
+func (d *Deployment) DeleteService(name string) error {
+	serviceSpec := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: d.namespace}}
+
+	d.logger.Infof("Deleting K8s service %s", name)
```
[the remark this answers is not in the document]
kfirtoledo: Usally K8s is with capital (https://kubernetes.io/)

pkg/deployment/k8s/deployment.go
elevran: Informers can be used and would work. I was suggesting to use a higher level interface (via the `controller-runtime` packages) but to make real use of it, it would run a control loop that is notified of Pod change events and then cache the needed object attributes on these changes. See, for example, the following example of using informers [here](https://github.com/elevran/podip) along with the needed code to cache objects (e.g., validate the Pod was assigned an IP, handle IPv4 and IPv6, etc.)

kfirtoledo: I put this code in different PR

cmd/cl-adm/templates/k8s.go
kfirtoledo: Done

pkg/deployment/k8s/deployment.go
kfirtoledo: Done

pkg/deployment/k8s/deployment.go
```diff
@@ -0,0 +1,227 @@
...
+	if err != nil {
+		d.logger.Errorf("error occurred while deleting K8s endpoint %s:", err)
+		return err
+	}
+
+	d.logger.Infof("Deleting K8s endPoint %s", name)
+	return nil
+}
+
+// GetPodLabelsByIP returns all the labels that match the pod IP.
```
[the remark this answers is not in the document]
kfirtoledo: In Different PR

kfirtoledo:
@elevran , @praveingk @orozery PTAL
1) I replace the deployment term with platform.
2) I created a reconciler that is responsible for handling  k8s objects (create, update, delete). In the future, I will add a reconcile function that will be responsible for k8s objects that failed in the deployment.
3) I removed the function for IP translation. I will add it later in a different PR

pkg/platform/k8s/reconciler.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		r.logger.Errorf("error occurred while deleting K8s %v %v:", reflect.TypeOf(obj).String(), err)
+		r.failedList[obj.GetName()] = reconcileObj{spec: obj, op: updateOp}
+		return
+	}
+
+	delete(r.list, obj.GetName())
+}
+
+// func (r *reconciler) reconcile() {
```
elevran, line 66: I think the current PR is fine as an **imperative** solution, that uses the controller-runtime clients. 
It is less of a classic reconciliation loop in k8s (which is what I had in mind initially), where the reconciler gets notified of changes in desired and/or actual state and then takes steps to have the actual (e.g, services, endpoints) match the desired (e.g., exports, imports) state.
I'm ok with making this change in a follow up PR

PR #75, opened 2023-10-10
Policy engine integration with new cp

zivnevo:
Includes:
* API calls to CRUD access policies
* Calling policy engine to authorize outgoing/incoming connections
* Policy engine sets remote peer for outgoing connections
* Calling policy engine to authorize adding bindings and exports
* Informing policy engine of added/removed peers and removed bindings and exports
* Persistent storage of access policies

Does not include:
* Handling load-balancing policies

### DIFF
--- a/cmd/gwctl/subcommand/policy.go
+++ b/cmd/gwctl/subcommand/policy.go
@@ -62,7 +62,7 @@ func (o *policyCreateOptions) run() error {
 	}
 	switch o.pType {
 	case policyengine.LbType:
-		return g.SendLBPolicy(o.serviceSrc, o.serviceDst, policyengine.PolicyLoadBalancer(o.policy), o.gwDest, client.Add)
+		return g.SendLBPolicy(o.serviceSrc, o.serviceDst, policyengine.LBScheme(o.policy), o.gwDest, client.Add)
 	case policyengine.AccessType:
 		policy, err := policyFromFile(o.policyFile)
 		if err != nil {
@@ -136,7 +136,7 @@ func (o *policyDeleteOptions) run() error {
 	}
 	switch o.pType {
 	case policyengine.LbType:
-		err = g.SendLBPolicy(o.serviceSrc, o.serviceDst, policyengine.PolicyLoadBalancer(o.policy), o.gwDest, client.Del)
+		err = g.SendLBPolicy(o.serviceSrc, o.serviceDst, policyengine.LBScheme(o.policy), o.gwDest, client.Del)
 	case policyengine.AccessType:
 		policy, err := policyFromFile(o.policyFile)
 		if err != nil {
--- a/cmd/gwctl/subcommand/policy.go
+++ b/cmd/gwctl/subcommand/policy.go
@@ -68,7 +68,13 @@ func (o *policyCreateOptions) run() error {
 		if err != nil {
 			return err
 		}
-		return g.SendAccessPolicy(policy, client.Add)
+		err = g.Policies.Create(policy)
+		if err != nil {
+			return err
+		}
+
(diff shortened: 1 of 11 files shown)


pkg/controlplane/store/types.go
```diff
@@ -101,3 +103,23 @@ func NewBinding(binding *api.Binding) *Binding {
...
+	api.Policy
+	// Version of the struct when object was created.
+	Version uint32
+}
+
+// Keys return the keys identifying the policy.
+func (ap *AccessPolicy) Keys() []string {
+	return []string{ap.Name}
+}
+
```
orozery, line 118:
```suggestion
```

This is actually a leftover from the time where a generic store existed.

pkg/policyengine/PolicyDispatcher.go
```diff
@@ -36,8 +39,27 @@ const (
...
+type PolicyDecider interface {
+	AddLBPolicy(lbPolicy *LBPolicy) error
+	DeleteLBPolicy(lbPolicy *LBPolicy) error
+
+	AddAccessPolicy(policy *api.Policy) error
+	DeleteAccessPolicy(policy *api.Policy) error
+
+	AuthorizeAndRouteConnection(connReq *event.ConnectionRequestAttr) (event.ConnectionRequestResp, error)
+
+	AddPeer(peer *api.Peer) error
```
orozery, line 51: These (and similar) functions cannot really fail.
Do we envision any future extension that will make them fail-able?

pkg/policyengine/PolicyDispatcher.go
```diff
@@ -36,8 +39,27 @@ const (
...
+	DeletePeer(name string) error
+
+	AddBinding(imp *api.Binding) (event.Action, error)
```
orozery, line 54: Currently this always returns "allow". Same for `AddExport`.
Do we plan to support policies disallowing this kind of operations?
@elevran @kfirtoledo

orozery (changes requested):

cmd/gwctl/subcommand/policy.go
```diff
@@ -182,26 +194,16 @@ func (o *policyGetOptions) run() error {
...
-	accessPolicies, err := g.GetAccessPolicies()
+	accessPolicies, err := g.Policies.List()
 	if err != nil {
 		return err
 	}
 
 	fmt.Printf("Access policies\n")
-	for d := range accessPolicies {
-		fmt.Printf("Access policy %d: %v\n", d, accessPolicies[d])
+	for d, policy := range *accessPolicies.(*[]api.Policy) {
```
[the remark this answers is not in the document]
zivnevo: The current code works.

The suggested change causes a panic
```
panic: interface conversion: interface {} is *[]api.Policy, not []*api.Policy
```
I copied this code from `cmd/gwctl/subcommand/import.go` and modified it to use `api.Policy`.

pkg/controlplane/store/types.go
zivnevo: Removed. Should I also remove the other implementations for the `Keys()` method?
```golang
func (exp *Export) Keys() []string
func (imp *Import) Keys() []string
func (b *Binding) Keys() []string
```

orozery: I've seen you already fixed bugs you found on the way (thanks!).
You can fix that as well, if you like.

pkg/policyengine/PolicyDispatcher.go
zivnevo: (this is something that was possible in the old policy engine, but is currently not implemented in the new policy engine)

orozery (approved):

PR #95, opened 2023-10-19
Delete support for old controlplane in policy engine

zivnevo:


orozery:
@kfirtoledo Have we decided it's time to remove the old controlplane/dataplane?

kfirtoledo:
Yes, that's what @elevran  said in the last sync meeting.

pkg/policyengine/PolicyDispatcher.go
```diff
@@ -50,7 +47,6 @@ const (
 )
```
kfirtoledo, line 47: Can you remove also MBG acronyms from the code.

kfirtoledo (changes requested):
You still need the accessControl.go or can you remove it

pkg/policyengine/PolicyDispatcher.go
zivnevo: Done

zivnevo:
> You still need the accessControl.go or can you remove it

Removed it. Good catch.

kfirtoledo (approved):

PR #96, opened 2023-10-19
tests/e2e: Add go dataplane to e2e-tests.

kfirtoledo:
Add go dataplane to the e2e tests.
Removing k8s.sh from the CI and use connectivity test with the go dataplane.


kfirtoledo:
Not needed for now

PR #97, opened 2023-10-19
Fix failing volume mount in "make copr-fix"

zivnevo:
Currently, when running `make copr-fix` I get (on both Ubuntu and Windows):
```
$ make copr-fix
adding copyright header...
docker run -it --rm -v :/github/workspace apache/skywalking-eyes header fix
docker: invalid spec: :/github/workspace: empty section between colons.
See 'docker run --help'.
make: *** [Makefile:48: copr-fix] Error 125
```

Seems like `$(pwd)` is not good enough on some environments.
Many suggest to replace it with `$(shell pwd)`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ vet-go: ; $(info vetting code...)
 	@go vet ./...
 
 copr-fix: ; $(info adding copyright header...)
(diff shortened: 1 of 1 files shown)


michalmalka:
LGTM

michalmalka (approved):

PR #99, opened 2023-10-19
cmd/cl-adm: Move core logic to a separate package

orozery:
This PR moves the core logic of the cl-adm command to a pkg/bootstrap. pkg/bootstrap implements a memory-based bootstrap, instead of the previous file-based approach which was implemented in cl-adm.
The cl-adm command still gives a file-based API, which is wrapped around the memory-based API that is implemented in pkg/bootstrap.

This PR is a prerequisite for a following PR for introducing a new e2e go-test framework.


kfirtoledo:
LGTM

cmd/cl-adm/cmd/create/create_peer.go
```diff
@@ -18,16 +18,13 @@ import (
...
-	"github.com/clusterlink-net/clusterlink/pkg/controlplane/api"
-	dpapi "github.com/clusterlink-net/clusterlink/pkg/dataplane/api"
-
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"golang.org/x/net/idna"
 
 	"github.com/clusterlink-net/clusterlink/cmd/cl-adm/config"
-	"github.com/clusterlink-net/clusterlink/cmd/cl-adm/util"
+	"github.com/clusterlink-net/clusterlink/pkg/bootstrap"
```
elevran, line 26: nit: Is `pkg/bootstrap` the right hierarchy? 
Feels like it is missing some context as to what we are bootstrapping...

elevran (approved):

cmd/cl-adm/cmd/create/create_peer.go
orozery: `pkg/bootstrap` contains code which generates certificates and configuration (e.g. k8s yaml) for deploying clusterlink.
I called it bootstrap as it is used only once in the lifecycle of a clusterlink site.
I'm open for other ideas :)

orozery: @elevran How about `pkg/deployment` / `pkg/deploy` ?

PR #103, opened 2023-10-19
Fix timing issue on go dataplane startup

praveingk:
This issue occurs when listenerFetcher is setup, while clusterFetcher is not ready yet, and a connection starts in the meantime. The problem is the long timeouts, and exiting the accept loop upon unable to find cluster.


pkg/dataplane/server/listener.go
```diff
@@ -69,7 +69,8 @@ func (d *Dataplane) serveEgressConnections(name string, listener net.Listener) e
 
 		targetHost, err := d.GetClusterHost(targetPeer)
 		if err != nil {
-			return err
+			d.logger.Errorf("Unable to get cluster host :%v.", err)
+			continue
```
orozery, line 73:
```suggestion
			conn.Close()
			continue
```

orozery (commented):
Can you comment out the go e2e test at the github action workflow?
Let's try and run it 5 times using the CI.

elevran (commented):
A suggestion for something else that can help mitigate:
Put the listener in pending state without opening the listening socket until there is a corresponding cluster (e.g., keep a pending listener map, keyed by cluster, check map on cluster updates).
The clusters can stills be removed later and result in a listener with no backing cluster (in which case you *could* close the listening socket and move it to pending again), so resetting a connection and logging the error is a good idea and probably the best fallback, regardless. 

P.S. the github e2e workflow commits could probably be done on a separate commit

praveingk:
@elevran This would require additional messages and coordination from the control-plane, since the dataplane doesn't know which listener are directed to which clusters. This decision comes from the control plane upon authorizing an incoming connection.  
Also, I think clusters aren't be removed from the dataplane without the knowledge of the control plane. Hence, controlplane should deny the connection if the corresponding cluster state (peer) isn't present in the controlplane.

elevran (approved):

Issue #86, opened 2023-10-16
Support IP translation K8s labels to enable policies on the source services.


kfirtoledo:
Support IP translation IP to pod labels for enablement of policies on the source services in the new control plane.

elevran:
Reference #17

kfirtoledo:
Done

PR #104, opened 2023-10-22
platform/k8s: Support GetLabelsFromIP

kfirtoledo:
Add controller-runtime to watch pod list and create ipToLabel list for
GetLabelFromIP function.


pkg/platform/k8s/pod_controller.go
elevran, line 1: `pod_reconciler.go` may be more accurate?

pkg/platform/k8s/pod_controller.go
```diff
@@ -0,0 +1,99 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// PodReconciler contain information on the clusters pods.
+type PodReconciler struct {
+	client.Client
+	ipToPod map[string]metav1.ObjectMeta
```
elevran, line 30: `ObjectMeta` probably contains more than you need (e.g., GKV, annotations, ....) - consider creating a local struct with only the fields we need (name, namespace, labels? maybe also ServiceAccount name)

elevran: Instead of `string` representation of the IP consider using `net/netip.Addr`?

pkg/platform/k8s/pod_controller.go
```diff
@@ -0,0 +1,99 @@
...
+	for _, ip := range pod.Status.PodIPs {
+		// ignoring host-networked Pod IPs
+		if ip.IP != pod.Status.HostIP {
+			r.ipToPod[ip.IP] = pod.ObjectMeta
+		}
+	}
+}
+
+// getLabelFromIP return all the labels for specific ip.
+func (r *PodReconciler) getLabelFromIP(ip string) map[string]string {
```
elevran, line 71: See earlier comment and retaining and returning more than just labels (e.g., name and namespace can also be used in the policies)

pkg/platform/k8s/pod_controller.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+	}
+}
+
+// updatePod adds or updates pod to ipToPod list.
+func (r *PodReconciler) updatePod(pod corev1.Pod) {
+	for _, ip := range pod.Status.PodIPs {
+		// ignoring host-networked Pod IPs
+		if ip.IP != pod.Status.HostIP {
+			r.ipToPod[ip.IP] = pod.ObjectMeta
```
elevran, line 65: nit: a more complex implementation (but with lower memory) would keep two maps:
`map[ip]namespacedName` and `map[namespacedName]podAttributes`. There can be multiple IPs pointing yo the same Pod (e.g. IPv4 and IPv6)

pkg/platform/k8s/pod_controller.go
elevran: I do not think we would get the IP address as a string from the data plane. It is more likely a `[]byte` which needs to be parsed. While you could delegate it to the caller, it might be better done here to keep a uniform interface to the Go and Envoy implementations.
You should, at least, document the string format (e.g., IPv4, v6, ...)

pkg/platform/k8s/pod_controller_test.go
```diff
@@ -0,0 +1,123 @@
...
+		logger:  logger,
+	}
+	req := ctrl.Request{NamespacedName: types.NamespacedName{
+		Name:      TestPodName,
+		Namespace: TestPodNameSpace,
+	}}
+
+	// Pod creation check
+	createLabel := "create-label"
+	pod := getFakePod(createLabel)
```
elevran, line 56: are you adding the Pod object to the fake client or creating it?
I'm not clear as to what the right approach would be, but I would expect the only methods exercised on the reconciler are the constructor and then reconcile() followed by lookup.

pkg/platform/k8s/pod_controller.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+	}
+}
+
+// getLabelFromIP return all the labels for specific ip.
```
elevran, line 70: Shouldn't this be an exported function? I expect it to be callable from the control plane...

elevran (changes requested):

elevran:
References #86

pkg/platform/k8s/pod_controller.go
kfirtoledo: Agree on the first comment.
About the second one I am not sure, the values for IP in the coreV1.Pod is also string and this is the one I am using and comparing to it.

pkg/platform/k8s/pod_controller.go
kfirtoledo: I think it is the responsibility of the controlplane to convert the bytes data to the correct IP format.

pkg/platform/k8s/pod_controller_test.go
kfirtoledo: I am trying to check just the reconcile logic, so I add or delete pods using the fake client and then call the reconcile to check the logic for adding pod info.

pkg/platform/k8s/pod_controller.go
kfirtoledo: There are two functions, the exported one is in the k8s/platform.go
I separated the communication between the components the controlplane uses only the functions of k8s/platform.go
and the platform will call al the functions related to the reconcilers.

zivnevo (commented):
LGTM (concentrated on the GetLabelsFromIP API)

zivnevo:
One question: Don't we need to extend the `Platform` **interface** as well, to include the `GetLabelsFromIP` method?
Seems like the controlplane instance holds a member of the abstract interface rather than the concrete `k8s.Platform`.

kfirtoledo:
@zivnevo you are right, see fix b3d4048d54996917f7b5eeb8c1c4db3c045d3388

elevran (approved):

PR #108, opened 2023-10-23
Run CI jobs in parallel; run CI jobs after merging to main

### LINKED ISSUE #100: Parallelize CI workflows
Currently, time from opening a PR to getting full testing results is at least 10 minutes. This makes fix-and-retest iterations rather slow. Furthermore, time to get even partial results is too long (e.g., unit tests do not run until lint is done). Splitting work to jobs and running these jobs in parallel can speed up this process.

In particular:
- [ ] In `.github/workflows/pr-check.yml`:  `unit-tests` doesn't have to be dependent on `static-checks-and-build`. Moreover, `make build` runs in both jobs, which is redundant.
- [ ] In `.github/workflows/pr-e2e-test.yml`: A first job can build docker images and upload them as an artifact. Then the 3 end-to-end tests can download this artifact and run their tests in parallel. See an example [here](https://github.com/docker/build-push-action/issues/225)

zivnevo:
Fixes #100 

Also, running checks after code is merged into `main`.


.github/workflows/pr-check.yml
```diff
@@ -1,6 +1,11 @@
...
 run-name: "PR #${{ github.event.number }} check (${{ github.sha }})"
-on: [pull_request]
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+
 jobs:
   static-checks-and-build:
```
orozery, line 10:
```suggestion
  static-checks:
```

orozery (approved):

.github/workflows/pr-check.yml
zivnevo: Fixed

PR #110, opened 2023-10-23
tests/e2e: Remove MBG acronym and old controlplane from e2e tests

kfirtoledo:
Remove support for old controlplane.
Remove MBG acronym from test folder.


tests/e2e/utils/kind.go
```diff
@@ -163,102 +163,46 @@ func DeleteCluster(name string) error {
...
+	err = runCmd("kind load docker-image cl-controlplane cl-dataplane cl-go-dataplane --name=" + name)
+	if err != nil {
+		return err
+	}
 
-		err = runCmd("kubectl create service nodeport dataplane --tcp=" + cPortLocal + ":" + cPortLocal + " --node-port=" + cPortStr)
-		if err != nil {
-			return err
-		}
+	err = runCmd("kubectl apply -f " + testOutputFolder + name + "/k8s.yaml")
```
elevran, line 184: ditto - can/should we use `client-go` directly? It can work with YAMLs as well

kfirtoledo: You are right, but most of this code will be replaced with the new framework so I will not fix it, I just wanted to start the cleanup and remove the code of the old controlplane.

PR #114, opened 2023-10-25
Loadbalancer cleanup before integration with new controlplane

zivnevo:
Related to #89 

* Removing http handlers supporting the old controlplane
* Improving error handling
* Improving LB resilience to bad paths (e.g., lookups before importing services)
* Unit-tests


pkg/policyengine/loadBalancer_test.go
```diff
@@ -0,0 +1,253 @@
...
+	// svc3 is imported from peer1 and peer2
+	lb.AddToServiceMap(svc1, peer1)
+	lb.AddToServiceMap(svc1, peer2)
+	lb.AddToServiceMap(svc2, peer1)
+	lb.AddToServiceMap(svc3, peer1)
+	lb.AddToServiceMap(svc3, peer2)
+}
+
+// We repeat lookups enough times to make sure we get all the peers allowed by the relevant policy
+func repeatLookups(t *testing.T, lb *policyengine.LoadBalancer, srcSvc, dstSvc string, peers []string) map[string]bool {
```
kfirtoledo, line 46: If it is only for valid peers, maybe you can stop earlier when all array is true

pkg/policyengine/loadBalancer.go
```diff
@@ -47,147 +45,120 @@ type ServiceState struct {
...
-	if _, ok := lB.Policy[serviceDst]; !ok { // Create default service if destination service is not exist
-		lB.Policy[serviceDst] = make(map[string]LBScheme)
+	if _, ok := lB.Scheme[serviceDst]; !ok { // Create default service if destination service is not exist
+		lB.Scheme[serviceDst] = make(map[string]LBScheme)
 	}
 	// start to update policy
-	lB.Policy[serviceDst][serviceSrc] = policy
+	lB.Scheme[serviceDst][serviceSrc] = scheme
+
 	if serviceDst != event.Wildcard { // ServiceStateMap[dst][*] is created only when the remote service is exposed
```
kfirtoledo, line 115: Need to fix the comment

kfirtoledo (approved):
LGTM, some small comments

pkg/policyengine/loadBalancer_test.go
zivnevo: Good idea. Will implement in next PR.

pkg/policyengine/loadBalancer.go
zivnevo: Right. Will fix in next PR.

PR #123, opened 2023-10-26
Loadbalancer integration into new controlplane

### LINKED ISSUE #89: Support for load-balancing policies in the new controlplane

zivnevo:
Fixes #89 

For now, using `api.Policy` to also pass load-balancing policies all the way from client to policy-engine (similar to access policies).
We may consider changing this later, by introducing a new API type.


pkg/policyengine/PolicyDispatcher.go
```diff
@@ -214,11 +214,29 @@ func connPolicyFromBlob(blob io.Reader) (*policytypes.ConnectivityPolicy, error)
...
+	lbPolicy := &LBPolicy{}
+	err := json.NewDecoder(blob).Decode(lbPolicy)
+	if err != nil {
+		plog.Errorf("failed decoding load-balancing policy: %v", err)
+		return nil, err
+	}
+	return lbPolicy, nil
+}
+
+func (pH *PolicyHandler) AddLBPolicy(policy *api.Policy) error {
```
kfirtoledo, line 227: Can you put a comment before those functions.

pkg/controlplane/instance.go
```diff
@@ -593,6 +657,7 @@ func NewInstance(peerTLS *util.ParsedCertData, storeManager store.Manager, platf
 		imports:       imports,
 		bindings:      bindings,
 		policies:      policies,
```
kfirtoledo, line 659: Maybe to change it to acPolicies

kfirtoledo (commented):
LGTM, some small comments

pkg/policyengine/PolicyDispatcher.go
zivnevo: Done

pkg/controlplane/instance.go
zivnevo: Done

kfirtoledo (approved):

PR #129, opened 2023-10-30
Wrap errors correctly

orozery:
This PR changes wrapped errors to use %w instead of %v.


tests/k8s.sh
elevran, line 1: nit: this could belong in a different PR which can be merged first to allow other (currently failing) PR's to pass.

elevran (approved):
question: is there a disabled linter that can be enabled once this merges?

PR #134, opened 2023-10-30
Group multiple dependencies in dependabot PRs

elevran:
By default, dependabot updates each dependency in its own PR. 
By grouping related packages (identified via specific package name patterns), we can create a single PR per group. 
This reduces the number of PRs as well as the effort in merging them.

The current groups are defined with relation to the dependencies listed in the `go.mod` file. Additional groups can be created in follow up PRs, if needed.


.github/dependabot.yml
zivnevo, line 1: Are you sure the `@` is needed? Seems to me like an npm thing.
I don't see it in other repos, e.g., https://github.com/cilium/certgen/blob/master/.github/dependabot.yml

zivnevo (changes requested):

zivnevo (approved):
Looks good

PR #139, opened 2023-10-30
Support new controlplane for iPerf3 Demo

kfirtoledo:
Split the demo functions into common.py, kind.py, and k8s.py files.
Fix the iperf3 test to support the new controlplane.


praveingk (approved):
Looks good. This can be complemented with a detailed readme with the CLI steps.

Issue #141, opened 2023-10-30
Go-dataplane taking too much time to establish xDS connection


orozery:
Attaching controlplane and dataplane logs (notice timestamps):
```
2023-10-30T12:27:41.427469735Z stderr F time="2023-10-30T12:27:41Z" level=info msg="Dataplane ID: f3c1077d-46d4-480e-b672-69a0617ac84c."
2023-10-30T12:27:41.427518574Z stderr F time="2023-10-30T12:27:41Z" level=info msg="Starting go dataplane, Name: peer1, ID: f3c1077d-46d4-480e-b672-69a0617ac84c"
2023-10-30T12:27:41.427972737Z stderr F time="2023-10-30T12:27:41Z" level=info msg="SNI proxy starting at :443." component=dataplane.server.http
2023-10-30T12:27:41.428004139Z stderr F time="2023-10-30T12:27:41Z" level=info msg="Creating listener on :443." component=listener name=sni-proxy
2023-10-30T12:27:41.428216721Z stderr F time="2023-10-30T12:27:41Z" level=info msg="Dataplane server starting at 127.0.0.1:8443." component=dataplane.server.http
2023-10-30T12:27:42.462691424Z stderr F time="2023-10-30T12:27:42Z" level=error msg="Failed to initialize type.googleapis.com/envoy.config.listener.v3.Listener fetcher: rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp 10.96.149.212:444: connect: connection refused\"." component=xds.client
2023-10-30T12:28:01.429533198Z stderr F time="2023-10-30T12:28:01Z" level=error msg="Failed to initialize type.googleapis.com/envoy.config.cluster.v3.Cluster fetcher: rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp 10.96.149.212:444: i/o timeout\"." component=xds.client
2023-10-30T12:28:01.508299088Z stderr F time="2023-10-30T12:28:01Z" level=info msg="Successfully initialized client for type.googleapis.com/envoy.config.cluster.v3.Cluster type." component=xds.client
2023-10-30T12:28:01.5083367Z stderr F time="2023-10-30T12:28:01Z" level=info msg="Starting to run type.googleapis.com/envoy.config.cluster.v3.Cluster fetcher." component=xds.client
[...]

PR #146, opened 2023-10-31
dataplane: Reduce xDS client connect timeout

orozery:
This PR reduces the xDS client connect timeout from the default 20 seconds to 1 second.
Fixes: #141.


elevran (approved):

praveingk (approved):
Good find!

PR #157, opened 2023-11-01
Enable and disable peers in the policy engine

### LINKED ISSUE #143: Inform the policy-engine of invalid peers
The heartbeat informs the control-plane which peer is active or not. The policy engine should be aware of it and not send it to a target that is not active. A simple solution is to add disablePeer() and enablePeer() functions to inform the policy-engine.

zivnevo:
Fixes #143 

`PolicyDispatcher` keeps track of the set of enabled peers, while `LoadBalancer` keeps track of all bindings.
When a call to `DisablePeer` is made, the peer is only deleted from the set of enabled peers in `PolicyDispatcher`, but `LoadBalancer` still maintains its bindings.
When a call to `EnablePeer` is made, the peer is added again to the set of enabled peers in `PolicyDispatcher`, and `LoadBalancer` still maintains its bindings.
When deciding on outgoing connections, `PolicyDispatcher` filters out peers that are not in the set of enabled peers before decisions are made at the Access PDP and at the Load Balancer.


pkg/policyengine/PolicyDispatcher.go
```diff
@@ -167,24 +161,31 @@ func (pH *PolicyHandler) AuthorizeAndRouteConnection(connReq *event.ConnectionRe
...
+
+	delete(pH.enabledPeers, name)
+	plog.Infof("Removed Peer %s", name)
+}
+
+func (pH *PolicyHandler) DisablePeer(name string) {
+	delete(pH.enabledPeers, name)
+}
+
+func (pH *PolicyHandler) EnablePeer(name string) {
```
kfirtoledo, line 200: Maybe you need to add check if it is exsit

kfirtoledo (changes requested):

pkg/policyengine/PolicyDispatcher.go
zivnevo: The [documentation](https://pkg.go.dev/builtin#delete)  says "If m is nil or there is no such element, delete is a no-op. "

pkg/policyengine/PolicyDispatcher.go
```diff
@@ -167,24 +161,31 @@ func (pH *PolicyHandler) AuthorizeAndRouteConnection(connReq *event.ConnectionRe
...
-	pH.knownPeers = append(pH.knownPeers[:index], pH.knownPeers[index+1:]...)
-	plog.Infof("Removed Peer(%s, %d) %+v", name, index, pH.knownPeers)
 
+	pH.loadBalancer.RemovePeerFromServiceMap(name)
+
+	delete(pH.enabledPeers, name)
+	plog.Infof("Removed Peer %s", name)
+}
+
+func (pH *PolicyHandler) DisablePeer(name string) {
```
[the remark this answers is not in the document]
zivnevo: It doesn't really matter. I treat `enabledPeers` as a set of strings. So `enabledPeers[peer]` will return `false` in both cases (because `false` is the 0-value of `bool`)

kfirtoledo (approved):

PR #158, opened 2023-11-01
controlplane/authz: Add source service to a connection request to the policy engine

kfirtoledo:
Add source service to a connection request to the policy engine.

### DIFF
--- a/pkg/controlplane/authz.go
+++ b/pkg/controlplane/authz.go
@@ -79,12 +79,18 @@ func (cp *Instance) AuthorizeEgress(req *EgressAuthorizationRequest) (*EgressAut
 		return nil, fmt.Errorf("no bindings found for import '%s'", req.Import)
 	}
 
-	// TODO: get k8s attributes using cp.kubeClient
 	connReq := eventmanager.ConnectionRequestAttr{DstService: req.Import, Direction: eventmanager.Outgoing}
+	srcLabels := cp.platform.GetLabelsFromIP(req.IP)
+	if src, ok := srcLabels["app"]; ok { // TODO: Add support for labels other than just the "app" key.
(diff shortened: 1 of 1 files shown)


praveingk (approved):

pkg/controlplane/authz.go
```diff
@@ -79,12 +79,20 @@ func (cp *Instance) AuthorizeEgress(req *EgressAuthorizationRequest) (*EgressAut
 		return nil, fmt.Errorf("no bindings found for import '%s'", req.Import)
 	}
 
-	// TODO: get k8s attributes using cp.kubeClient
 	connReq := eventmanager.ConnectionRequestAttr{DstService: req.Import, Direction: eventmanager.Outgoing}
+	srcLabels := cp.platform.GetLabelsFromIP(req.IP)
+
+	if srcLabels != nil {
```
zivnevo, line 85: no need to check for nil value. [The nil map supports queries](https://go.dev/ref/spec#Index_expressions), and will return the same results as an empty map.
It's only insertions to the nil map that panic.

zivnevo (approved):
Just these two nits

pkg/controlplane/authz.go
```diff
@@ -79,12 +79,20 @@ func (cp *Instance) AuthorizeEgress(req *EgressAuthorizationRequest) (*EgressAut
...
+		if src, ok := srcLabels["app"]; ok {
+			connReq.SrcService = src
+		}
+	}
```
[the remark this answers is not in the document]
kfirtoledo: Done

pkg/controlplane/authz.go
kfirtoledo: Done

PR #162, opened 2023-11-05
Support for dev-container

elevran:
Changed Makefile targets and added script to install dev tools

Signed-off-by: Etai Lev Ran <elevran@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ prereqs-force: ; $(info force installing dev tooling...)
 	./hack/install-devtools.sh --force
 
 .dev-container: Containerfile.dev
-	docker build -f Containerfile.dev -t $(IMG) .
+	docker build -f Containerfile.dev -t quay.io/$(IMAGE_ORG)/dev:latest .
 	touch $@
 
 .PHONY: run-dev-container
@@ -33,7 +33,7 @@ run-dev-container: .dev-container
 		-v /var/run/docker.sock:/var/run/docker.sock \
 		-v $(CURDIR):$(CURDIR) \
 		--workdir $(CURDIR) \
-		$(IMG)
+		quay.io/$(IMAGE_ORG)/dev:latest
 
 
 #-- precommit code checks --
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,19 @@ prereqs: ; $(info installing dev tooling...)
 prereqs-force: ; $(info force installing dev tooling...)
 	./hack/install-devtools.sh --force
 
+.dev-container: Containerfile.dev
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -14,18 +14,29 @@ clean: ; $(info cleaning previous builds...)	@
...
+#-- development tooling --
+.PHONY: prereqs prereqs-force
 
+prereqs: ; $(info installing dev tooling...) 
+	./hack/install-devtools.sh
+
+prereqs-force: ; $(info force installing dev tooling...)
+	./hack/install-devtools.sh --force
+
+.dev-container: Containerfile.dev
```
kfirtoledo, line 26: Containerfile.dev should be removed?

kfirtoledo (changes requested):

Makefile
elevran: the creation of the devcontainer should repeat itself when the definition file changes.
What is the reason to remove the definition file from being a dependency?

Makefile
```diff
@@ -14,18 +14,29 @@ clean: ; $(info cleaning previous builds...)	@
...
+	docker build -f Containerfile.dev -t $(IMG) .
+	touch $@
+
+.PHONY: run-dev-container
+run-dev-container: .dev-container
```
kfirtoledo, line 31: What is the motivation to create a container with all the dependencies but without the clusterlink images

Makefile
```diff
@@ -14,18 +14,29 @@ clean: ; $(info cleaning previous builds...)	@
...
+	docker run --rm -it --network bridge \
+		-v /var/run/docker.sock:/var/run/docker.sock \
+		-v $(CURDIR):$(CURDIR) \
```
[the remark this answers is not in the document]
elevran: It is an automatic `make` variable, not a shell variable.

From Make manual:
```
For your convenience, when GNU make starts (after it has processed any -C options) it sets the variable CURDIR to the pathname of the current working directory.
```

Makefile
elevran: It is a container that can be used to develop in isolation (i.e., without installing anything into your machine).
The [`CodeSpaces`](https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/adding-a-dev-container-configuration/introduction-to-dev-containers) button on GitHub works with it, [`vscode`](https://code.visualstudio.com/docs/devcontainers/containers) can create the container for you, etc.

The only additional step is mounting the ClusterLink source code into the container so it can be edited, built, etc. This is done via the Makefile (vscode does it automatically, see above)

kfirtoledo (approved):

PR #164, opened 2023-11-05
Move the utils function for the demos to be under the cluster class.

kfirtoledo:
Move the utils function for the demos to be under the cluster class,
for reusability, and supports the same demo in the kind and cloud environment.


elevran (changes requested):
- there seems to be some missing newlines at end of files
- how does this relate to `iperf` changes in #170?

demos/bookinfo/kind/test.py
```diff
@@ -23,133 +23,30 @@
...
-   runcmd(f'kubectl exec -i {gwctl3Pod} -- gwctl create policy --type access --policyFile {policyFile}')
-
-   print(f"Proctpage1 url: http://{gw1Ip}:30001/productpage")
-   print(f"Proctpage2 url: http://{gw1Ip}:30002/productpage")
+   gw1           = cluster(name="peer1")
+   gw2           = cluster(name="peer2")
+   gw3           = cluster(name="peer3")
 
+   bookInfoDemo(gw1, gw2, gw3, testOutputFolder, args["logLevel"], args["dataplane"])
+   print(f"Proctpage1 url: http://{gw1.ip}:30001/productpage")
```
praveingk, line 50: Typo productpage1

praveingk (changes requested):

demos/bookinfo/kind/test.py
kfirtoledo: Done

kfirtoledo:
@elevran,  The main purpose of this change is that the same test (most of the common parts) will be run in kind and the cloud so that it will be easy to develop the demo in kind and then check it on the cloud.

kfirtoledo:
@praveingk, @elevran PTAL

praveingk (approved):

PR #170, opened 2023-11-07
Demos: update support to cloud environment. 

kfirtoledo:
Move all demo functions to the cloud cluster class and update the iPerf3 test to run on GCP.


demos/utils/cloud.py
```diff
@@ -0,0 +1,156 @@
...
+        gwIp=""
+        if externalIp !="":
+            runcmd("kubectl patch svc cl-dataplane-load-balancer -p "+ "\'{\"spec\": {\"type\": \"LoadBalancer\", \"loadBalancerIP\": \""+ externalIp+ "\"}}\'")
+            gwIp= externalIp
+        while gwIp =="":
+            print("Waiting for cl-dataplane ip...")
+            gwIp=sp.getoutput('kubectl get svc -l app=cl-dataplane  -o jsonpath="{.items[0].status.loadBalancer.ingress[0].ip}"')
+            time.sleep(10)
+        self.ip = gwIp
+        return gwIp
```
elevran, line 156: nit: missing newline at EoF

elevran (commented):
A few clarifications:
- where are credentials coming from? I assume that the user is expected to log-in to the respective cloud and then the cloud specific CLI picks up the access token from the file system.
- how do you monitor the test to ensure that no resources (VPC, clusters, IP addresses...) remain allocated after the test completes? For success you can call delete on those, but how do we handle failure cases?
- can more than one instance of the test be run concurrently or would they interfere?

demos/utils/cloud.py
```diff
@@ -0,0 +1,158 @@
...
+        self.name        = name
+        self.zone        = zone
+        self.platform    = platform
+        self.machineType = machineType
+        self.port        = 443
+        self.ip          = ""
+        self.nodeIP      = "" # the IP of the worker node
+    
+    # createCluster creates a K8s cluster on cloud platform.
+    def createCluster(self, runBg):
```
praveingk, line 32: Can we have a method to setCluster parameters, if the cluster is already present and someone wants to test deployment, without having to recreate cluster?

demos/iperf3/cloud/test.py
```diff
@@ -11,121 +12,80 @@
...
-
-    #Test MBG1
-    connectToCluster(mbg1)
-    podIperf3= getPodName(srcSvc)
-    mbgPod,mbgPodIP=getPodNameIp("mbg")
-    mbg1LocalPort, mbg1ExternalPort = getMbgPorts(mbgPod,destSvc)
     for i in range(2):
         printHeader(f"iPerf3 test {i}")
-        cmd = f'kubectl exec -i {podIperf3} --  iperf3 -c {mbgPodIP} -p {mbg1LocalPort} -t 40'
+        cmd = f'kubectl exec -i {podIperf3} --  iperf3 -c iperf3-server -p {5000} -t 40'
```
praveingk, line 90: Do we have a cleanup script to delete the clusters? or its in the Makefile?

praveingk (changes requested):
Can we have a README which explains the below :
1) Login method to cloud, and how the demo script manages.
2) How the certificates are created and dispersed to the gateways.
3) Other details as in the kind setup.

demos/utils/cloud.py
kfirtoledo: Yes I split it to createCluster and startCluster

demos/iperf3/cloud/test.py
kfirtoledo: Not a script but the test parameter you can decide if you want to clean or delete the clusters but you need to do it manually.

kfirtoledo:
@elevran, @praveingk, From my point of view, the purpose of all this demo folder( after we move the test checking to go) is that we will have an automated script to present demos in kind or in a cloud environment. That's why the script not delete the clusters at the end of the tests (it can be done by running the same script again with -c delete or -c  clean just to remove the deployment, services and claims).
I will create another readme file that will explain how to set Clusterlink in the cloud environment or in-kind environment using cl-adm. I will add the link to the iPerf3 readme tutorial (it is now under demo, but we can move it to a new tutorials folder) about credentials and how to create a k8s cluster in GCP and connect to it, I can add it to the readme, but I don't think it should be part of it, because we assume the user already create the k8s cluster.

praveingk (approved):

PR #171, opened 2023-11-07
Add `-v` / `--version` to executables 

### LINKED ISSUE #169: Add version strings to executables
Add build/version information to executables (e.g., data and control planes, gwctl, etc.)
Information can include:
- git SHA
- build date
- git dirty/clean
- official version (e.g. SemVer values)

elevran:
Version information is automatically generated from go build information.

Fixes #169


pkg/versioninfo/variables.go
```diff
@@ -0,0 +1,46 @@
...
+// Unfortunately Go does not allow ".." in embed paths. A possible workaround is to
+// use "go:generate" comment to copy the file from the root to this location (and adding
+// the local version to .gitignore...)
+// An alternative would be to use a go:generate comment with "git describe" to find
+// the latest tag reachable from the current revision..
+var (
+	//go:embed version.txt
+	semver string
+	// SemVer will be the version specified in version.txt file
+	SemVer = strings.TrimSpace(semver)
```
kfirtoledo, line 38: You don't prefer to do it in the init functions?

pkg/versioninfo/version.go
```diff
@@ -0,0 +1,66 @@
...
+
+// Short provides a short string summarizing available version information.
+// The format is <SemVer>-<GIT SHA>[-<dirty>]
+func Short() string {
+	parts := make([]string, 0, 4)
+
+	if SemVer != "" {
+		parts = append(parts, SemVer)
+	}
+	if Version != "unknown" && Version != "(devel)" {
```
kfirtoledo, line 47: I didn't saw in the code where you update the version

pkg/versioninfo/variables.go
elevran: It is not needed - this way it is done at build time, not runtime.

pkg/versioninfo/version.go
elevran: Version will be the version tag if the binary is built with "go install url/package@version". If the binary is built some other way, it will be "(devel)".
So in our case you're correct - it'll remain as "devel"/"unknown". I can handle it in a follow up PR based on the TODO in varaibles.go

kfirtoledo (approved):

PR #173, opened 2023-11-08
nit fixes

orozery:
1. Change log output directory for k8s.sh
2. Change gwctl pod to trust only local peer.


tests/k8s.sh
```diff
@@ -27,10 +27,11 @@ function clean_up {
```
kfirtoledo, line 27: We don't want to remove this file?

kfirtoledo (approved):

PR #176, opened 2023-11-09
cmd/cl-adm: Add container registry flag

kfirtoledo:
Add a remote-image flag that chooses whether to use a local or remote image. 
Add a container-registry flag that represents the container registry to pull the project images.

### DIFF
--- a/cmd/cl-adm/cmd/create/create_peer.go
+++ b/cmd/cl-adm/cmd/create/create_peer.go
@@ -37,6 +37,8 @@ type PeerOptions struct {
 	DataplaneType string
 	// LogLevel is the log level.
 	LogLevel string
+	// ContainerRegistry is the container registry to pull the project images.
+	ContainerRegistry string
 }
 
 // AddFlags adds flags to fs and binds them to options.
@@ -44,8 +46,8 @@ func (o *PeerOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.Name, "name", "", "Peer name.")
 	fs.Uint16Var(&o.Dataplanes, "dataplanes", 1, "Number of dataplanes.")
 	fs.StringVar(&o.DataplaneType, "dataplane-type", platform.DataplaneTypeEnvoy, "Type of dataplane, Supported values: \"envoy\" (default), \"go\"")
-	fs.StringVar(&o.LogLevel, "log-level", "info",
-		"The log level. One of fatal, error, warn, info, debug.")
+	fs.StringVar(&o.LogLevel, "log-level", "info", "The log level. One of fatal, error, warn, info, debug.")
+	fs.StringVar(&o.ContainerRegistry, "container-registry", "ghcr.io/clusterlink-net", "The container registry to pull the project images. If empty will use local registry.")
 }
 
 // RequiredFlags are the names of flags that must be explicitly specified.
@@ -190,6 +192,7 @@ func (o *PeerOptions) Run() error {
 		Dataplanes:              o.Dataplanes,
 		DataplaneType:           o.DataplaneType,
 		LogLevel:                o.LogLevel,
+		ContainerRegistry:       o.ContainerRegistry,
 	})
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


cmd/cl-adm/cmd/create/create_peer.go
```diff
@@ -37,15 +37,20 @@ type PeerOptions struct {
 	DataplaneType string
 	// LogLevel is the log level.
 	LogLevel string
+	// RemoteImage determines whether to use a local image or pull it from a container registry.
+	RemoteImage bool
```
orozery, line 41:
```suggestion
```

Let's remove this argument.
An empty `ContainerRegistry` will denote a local image. All other values denote a remote image.

pkg/bootstrap/platform/k8s.go
```diff
@@ -156,7 +156,7 @@ spec:
             secretName: cl-dataplane
       containers:
         - name: dataplane
-          {{ if (eq .dataplaneType .dataplaneTypeEnvoy) }}image: cl-dataplane{{ else }}image: cl-go-dataplane{{ end }}
+          {{ if (eq .dataplaneType .dataplaneTypeEnvoy) }}image: {{.containerRegistry}}cl-dataplane{{ else }}image: {{.containerRegistry}}cl-go-dataplane{{ end }}
```
orozery, line 159:
```suggestion
          image: {{.containerRegistry}}{{ if .containerRegistry }}/{{ end }}{{ if (eq .dataplaneType .dataplaneTypeEnvoy) }}cl-dataplane{{ else }}cl-go-dataplane{{ end }}
```

pkg/bootstrap/platform/k8s.go
```diff
@@ -191,7 +191,7 @@ spec:
         secretName: gwctl
   containers:
     - name: gwctl
-      image: gwctl
+      image: {{.containerRegistry}}gwctl
```
orozery, line 194:
```suggestion
      image: {{.containerRegistry}}{{ if .containerRegistry }}/{{ end }}gwctl
```

orozery (changes requested):

cmd/cl-adm/cmd/create/create_peer.go
kfirtoledo: I think we need to keep him because most of the users will use the remote registry and it is much easier to add the flag  --remote-image instead of  container-registry=ghcr.io/clusterlink-net/

orozery: Are we planning to release our images to `gchr.io`?
In that case, let's set the default registry to gchr, and the user will not have to use any flag.
For local images, user will have to set `containerRegistry` to an empty string (which of course will be documented).

kfirtoledo: Ok, by the way the images are already in the ghcr : https://github.com/orgs/clusterlink-net/packages

pkg/bootstrap/platform/k8s.go
kfirtoledo: I prefer to do it in the go code, see my next patch.

cmd/cl-adm/cmd/create/create_peer.go
```diff
@@ -37,15 +37,19 @@ type PeerOptions struct {
...
```
orozery, line 41: Forgot to remove this

pkg/bootstrap/platform/k8s.go
orozery, line 159:
```suggestion
          image: {{.containerRegistry}}{{ if (eq .dataplaneType .dataplaneTypeEnvoy) }}cl-dataplane{{ else }}{{.containerRegistry}}cl-go-dataplane{{ end }}
```

pkg/bootstrap/platform/k8s.go
```diff
@@ -269,11 +269,16 @@ subjects:
 
 // K8SConfig returns a kubernetes deployment file.
 func K8SConfig(config *Config) ([]byte, error) {
+	if config.ContainerRegistry != "" {
+		config.ContainerRegistry += "/"
```
orozery, line 273: This will mutate the caller struct, which IMO is not desirable.
Can you use a local variable instead?

orozery (changes requested):

orozery (approved):

PR #178, opened 2023-11-12
controlplane: connect enable/disable peer to heartbeat

kfirtoledo:
Update peer status in policy-engine according to the heartbeat status.

### DIFF
--- a/pkg/controlplane/client.go
+++ b/pkg/controlplane/client.go
@@ -39,12 +39,13 @@ const (
 // client for accessing a remote peer.
 type client struct {
 	// jsonapi clients for connecting to the remote peer (one per each gateway)
-	clients    []*jsonapi.Client
-	lastSeen   time.Time
-	active     bool
-	stopSignal chan struct{}
-	lock       sync.RWMutex
-	logger     *logrus.Entry
+	clients            []*jsonapi.Client
+	lastSeen           time.Time
+	active             bool
+	stopSignal         chan struct{}
+	lock               sync.RWMutex
+	logger             *logrus.Entry
+	peerStatusCallback func(bool) // Callback function for notifying changes in peer
 }
 
 // remoteServerAuthorizationResponse represents an authorization response received from a remote controlplane server.
@@ -114,10 +115,16 @@ func (c *client) IsActive() bool {
 // setActive the peer status (active or not).
 func (c *client) setActive(active bool) {
 	c.lock.Lock()
-	defer c.lock.Unlock()
+	activePrevState := c.active
 	c.active = active
 	if active || c.lastSeen.IsZero() {
 		c.lastSeen = time.Now()
+	}
+	c.lock.Unlock()
+
+	// Update other components like the policy engine with the peer status.
+	if active != activePrevState && c.peerStatusCallback != nil {
+		c.peerStatusCallback(active)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/controlplane/instance.go
```diff
@@ -87,6 +87,15 @@ func (cp *Instance) CreatePeer(peer *cpstore.Peer) error {
 
 	cp.policyDecider.AddPeer(&api.Peer{Name: peer.Name, Spec: peer.PeerSpec})
 
+	client.SetPeerStatusCallback(func(status bool) {
```
orozery, line 90:
```suggestion
	client.SetPeerStatusCallback(func(isActive bool) {
```

pkg/controlplane/instance.go
```diff
@@ -87,6 +87,15 @@ func (cp *Instance) CreatePeer(peer *cpstore.Peer) error {
...
+		if status {
```
orozery, line 91:
```suggestion
		if isActive {
```

pkg/controlplane/client.go
```diff
@@ -113,11 +114,23 @@ func (c *client) IsActive() bool {
...
+		c.lock.Lock()
+		defer c.lock.Unlock()
+		activePrevState = c.active
+		c.active = active
+		if active || c.lastSeen.IsZero() {
+			c.lastSeen = time.Now()
+		}
+	}
+	// Update other components like the policy engine with the peer status.
+	if !activePrevState && active {
```
elevran, line 128:
```suggestion
	if active != activePrevState {
	   c.c.peerStatusCallback(active)
	}
```
can be an alternative to L128-134

pkg/controlplane/client.go
```diff
@@ -113,11 +114,23 @@ func (c *client) IsActive() bool {
...
+		c.peerStatusCallback(true)
+	}
+
+	if activePrevState && !active {
+		c.peerStatusCallback(false)
 	}
```
orozery, line 134:
```suggestion
	if activePrevState != active && c.peerStatusCallback != nil {
		c.peerStatusCallback(active)
	}
```

elevran (approved):

orozery (changes requested):

pkg/controlplane/client.go
```diff
@@ -113,11 +114,19 @@ func (c *client) IsActive() bool {
...
 func (c *client) setActive(active bool) {
-	c.lock.Lock()
-	defer c.lock.Unlock()
-	c.active = active
-	if active || c.lastSeen.IsZero() {
-		c.lastSeen = time.Now()
+	var activePrevState bool
+	{
+		c.lock.Lock()
+		defer c.lock.Unlock()
```
orozery, line 120: I don't think that a block helps to make this defer run when you want it.
i.e. https://stackoverflow.com/questions/49456943/why-is-gos-defer-scoped-to-function-not-lexical-enclosure
Let's revert this block, and instead call unlock directly at the end of it.

orozery (changes requested):

pkg/controlplane/client.go
```diff
@@ -113,11 +114,20 @@ func (c *client) IsActive() bool {
...
-	c.active = active
-	if active || c.lastSeen.IsZero() {
-		c.lastSeen = time.Now()
+	{
+		activePrevState = c.active
+		c.active = active
+		if active || c.lastSeen.IsZero() {
+			c.lastSeen = time.Now()
+		}
+	}
```
orozery, line 125:
```suggestion
	activePrevState := c.active
	c.active = active
	if active || c.lastSeen.IsZero() {
		c.lastSeen = time.Now()
	}
```

And remove `var activePrevState bool`

orozery (changes requested):

orozery (approved):

PR #182, opened 2023-11-13
tests/e2e/k8s: Allow timeout errors

orozery:
This PR fixes the function that waits for the controlplane server to come up, to allow timeout errors. This should resolve CI errors that mention: context deadline exceeded (Client.Timeout exceeded while awaiting headers).
e.g. https://github.com/clusterlink-net/clusterlink/actions/runs/6839437393/job/18597588234


kfirtoledo (approved):

kfirtoledo:
Looks good, but we probably need a better way to know if the control plane is up.

PR #198, opened 2023-11-16
pkg/platform: Fix update and external service for k8s platform

kfirtoledo:
Fix update for k8s object and add update to k8s rules.
Save k8s object by name in the reconcile and not by host.
Fix external service by creating an external service instead of an endpoint.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -74,8 +74,8 @@ type ExportSpec struct {
 	Service Endpoint
 	// ExternalService specifies a service located outside of the cluster.
 	// If ExternalService is defined, a new local ExternalName Service will be created,
-	// pointing to the ExternalService. Note that the port value is the same as defined
-	// in the Service field.
+	// pointing to the ExternalService. Note: the external name is just a DNS name mapping
+	// to the Service field, and there is no associated port mapping.
 	ExternalService string
 }
 
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -72,10 +72,11 @@ type ExportSpec struct {
 	// have a different `Name`. The endpoint Host name could be different from Name,
 	// decoupling service names between Peers.
 	Service Endpoint
-	// ExternalService endpoint defines an endpoint located outside of the cluster.
-	// When this endpoint is defined, a new local Endpoint will be created.
-	// This local Endpoint serves as a bridge, directing the Service to the ExternalService.
-	ExternalService Endpoint
+	// ExternalService specifies a service located outside of the cluster.
+	// If ExternalService is defined, a new local ExternalName Service will be created,
+	// pointing to the ExternalService. Note that the port value is the same as defined
+	// in the Service field.
+	ExternalService string
 }
 
 // Import defines a service that is being imported to the local Peer from a remote Peer.
--- a/pkg/controlplane/instance.go
+++ b/pkg/controlplane/instance.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"sync"
 
+	valid "github.com/asaskevich/govalidator"
 	"github.com/envoyproxy/go-control-plane/pkg/cache/v3"
 	"github.com/lestrrat-go/jwx/jwk"
 	"github.com/sirupsen/logrus"
@@ -34,6 +35,7 @@ import (
 
 const (
 	dataplaneAppName = "cl-dataplane"
+	exportPrefix     = "export_"
 )
 
 // Instance of a controlplane, where all API servers delegate their requested actions to.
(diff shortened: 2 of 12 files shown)


pkg/api/types.go
```diff
@@ -72,10 +72,11 @@ type ExportSpec struct {
...
 	// decoupling service names between Peers.
 	Service Endpoint
-	// ExternalService endpoint defines an endpoint located outside of the cluster.
-	// When this endpoint is defined, a new local Endpoint will be created.
-	// This local Endpoint serves as a bridge, directing the Service to the ExternalService.
-	ExternalService Endpoint
+	// ExternalService specifies a service located outside of the cluster.
+	// If ExternalService is defined, a new local ExternalName Service will be created,
+	// pointing to the ExternalService. Note that the port value is the same as defined
+	// in the Service field.
```
elevran, line 78: Note: the external name is **just a DNS name mapping** (i.e., when some workload resolves "servicename.namespace.cluster.local" CoreDNS return the IP address associated with "externalname").
Unlike regular services, there is **no port mapping** involved and no Service VIP allocated.

pkg/controlplane/instance.go
```diff
@@ -166,11 +168,10 @@ func (cp *Instance) GetAllPeers() []*cpstore.Peer {
 // CreateExport defines a new route target for ingress dataplane connections.
 func (cp *Instance) CreateExport(export *cpstore.Export) error {
 	cp.logger.Infof("Creating export '%s'.", export.Name)
-	exSvc := export.ExportSpec.ExternalService
-	if (exSvc.Host != "" && exSvc.Port == 0) || (exSvc.Host == "" && exSvc.Port != 0) {
-		return fmt.Errorf("ExternalService (Host: %s ,Port: %d) wasn't set properly", exSvc.Host, exSvc.Port)
+	eSpec := export.ExportSpec
+	if eSpec.ExternalService != "" && !valid.IsIP(eSpec.ExternalService) && !valid.IsDNSName(eSpec.ExternalService) {
```
elevran, line 172: nit: is there a way to validate these without adding a package? For example, [net.ParseIP](https://pkg.go.dev/net#example-ParseIP) and actually resolving the DNS name if it's not an IP?

pkg/controlplane/instance.go
```diff
@@ -184,10 +185,9 @@ func (cp *Instance) CreateExport(export *cpstore.Export) error {
 		if err := cp.exports.Create(export); err != nil {
 			return err
 		}
-		// create k8s endpoint and service for external service.
-		if exSvc.Host != "" && exSvc.Port != 0 && cp.initialized {
-			cp.platform.CreateEndpoint(export.Name, exSvc.Host, exSvc.Port)
-			cp.platform.CreateService(export.Name, export.Name, exSvc.Port, exSvc.Port)
+		// create a k8s external service.
+		if eSpec.ExternalService != "" {
+			cp.platform.CreateExternalService(exportPrefix+export.Name, eSpec.Service.Host, eSpec.ExternalService, eSpec.Service.Port)
```
elevran, line 190: See earlier comment regarding the port - not sure it is needed.
For example, there is no use for port in `setExternalNameService`.

elevran (changes requested):
mostly around use of port (in comments and code) for ExternalName service

pkg/controlplane/instance.go
kfirtoledo: Yes the port it is just for the print, I can remove it

pkg/api/types.go
kfirtoledo: Check now if it is more clear.

pkg/controlplane/instance.go
kfirtoledo: Done, use an internal function.

pkg/platform/k8s/platform.go
```diff
@@ -18,6 +18,7 @@ import (
 	"os"
 
 	logrusr "github.com/bombsimon/logrusr/v4"
+	"github.com/clusterlink-net/clusterlink/pkg/utils/netutils"
```
elevran, line 21: import order seems off. Does it pass the `make precommit` check?

pkg/platform/platform.go
```diff
@@ -15,11 +15,10 @@ package platform
...
-	UpdateService(name, targetApp string, port, targetPort uint16)
-	DeleteService(name string)
-	CreateEndpoint(name, targetIP string, targetPort uint16)
-	UpdateEndpoint(name, targetIP string, targetPort uint16)
-	DeleteEndpoint(name string)
+	CreateService(name, host, targetApp string, port, targetPort uint16)
+	UpdateService(name, host, targetApp string, port, targetPort uint16)
+	DeleteService(name, host string)
+	CreateExternalService(name, host, ExternalName string)
+	UpdateExternalService(name, host, ExternalName string)
```
elevran, line 22: confirming: there's no corresponding `DeleteExternalService` since it uses the `DeleteService` call, righ?

elevran (changes requested):
please fix import order. Other comments are nits

pkg/platform/platform.go
kfirtoledo: Yes

pkg/platform/k8s/platform.go
kfirtoledo: Yes and this is the order that the compiler enforces (I can't change it).

PR #211, opened 2023-11-21
controlplane: Make policies deletable by name

orozery:
This PR changes the API for deleting policies (both access and LB). Instead of deleting by value, delete by name.


zivnevo (approved):
LGTM

PR #212, opened 2023-11-21
policyengine: Don't remove bindings of a deleted peer

orozery:
This {R changes the policy dispatcher to avoid deletion of bindings relevant to a deleted peer.
The peer is still removed from the enabled peers list.
This allows for re-creating a deleted peer without re-creating the relevant bindings.

### DIFF
--- a/pkg/controlplane/instance.go
+++ b/pkg/controlplane/instance.go
@@ -85,15 +85,15 @@ func (cp *Instance) CreatePeer(peer *cpstore.Peer) error {
 		return err
 	}
 
-	cp.policyDecider.AddPeer(&api.Peer{Name: peer.Name, Spec: peer.PeerSpec})
+	cp.policyDecider.AddPeer(peer.Name)
 
 	client.SetPeerStatusCallback(func(isActive bool) {
 		if isActive {
-			cp.policyDecider.EnablePeer(peer.Name)
+			cp.policyDecider.AddPeer(peer.Name)
 			return
 		}
 
-		cp.policyDecider.DisablePeer(peer.Name)
+		cp.policyDecider.DeletePeer(peer.Name)
 	})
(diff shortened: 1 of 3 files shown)


zivnevo:
I don't get it. I thought this is exactly the reason why we added `EnablePeer()` and `DisablePeer()` to the interface.

orozery:
> I don't get it. I thought this is exactly the reason why we added `EnablePeer()` and `DisablePeer()` to the interface.

`EnablePeer` and `DisablePeer()` are called according to periodic monitoring (healthchecks) of peers.

The change proposed here is for the function that is called when the user explicitly deletes a peer. In this case the peer is removed from the controlplane store, and its periodic monitoring is stopped.
From the policy engine perspective, the peer is removed from the list of enabled peers (same as if healthcheck failed and `DisablePeer` was called).
However, the bindings which point to this peer should not be deleted, as they are independent objects.
For example, you can create a binding to a non-existing peer, which is OK.
The policy engine will never actually use this binding until the peer is "enabled" (by first being created, then healthchecked).

zivnevo:
ok, but now `DeletePeer` and `DisablePeer` have pretty much the same code. Same for `AddPeer` and `EnablePeer`. Perhaps we don't need this duplication? If bindings of deleted peers are kept, then maybe healthcheck can simply call `AddPeer` and `DeletePeer` and we can simplify the interface...
@kfirtoledo , what do you think?

kfirtoledo:
Sounds ok, If we want to support an option for binding to non-existing peers, I didn't know this option is legal.

orozery:
@zivnevo I've implemented your suggestion and removed `EnablePeer` and `DisablePeer`.

zivnevo (approved):
LGTM

PR #221, opened 2023-11-29
Design proposal for automating deployment via operator and CRD

kfirtoledo:
Design proposal for ClusterLink deployment using an operator.
Reference #286


design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+    git clone git@github.com:clusterlink-net/clusterlink.git
+
+Install the ClusterLink CLI:
+
+    cd clusterlink
+    make install
+
+In the future, these two commands should be replaced the command:
+
+    curl -L https://clusterlink.net/install.sh | sh
```
elevran, line 47: Or GH release or helm or ...

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Create Fabric certificate
```
elevran, line 49: I would structure it differently. Assuming we have a "deployment" CRD that the cluster administrator interacts with, I would suggest specifying 
- prerequisites (e.g., certificates as inputs, who has access to them, where do they come from, are CRD and operator already defined and running on the cluster, etc.)
- CRD format (what fields, possible values, explanation, etc.)
- outputs based on the CRDs (what objects are created, kept in sync, etc,)

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+Create the fabric certificate, which is the root CA certificate for all the K8s clusters:
+
+    clusterlink create fabric --name <fabric_name>
```
elevran, line 52: I see the following commands run by a *fabric administrator*:
- `clusterlink create fabric <fabric_name>` - produces CA certificate and key 
- `clusterlink create site --fabric <fabric_name> --name <site_name>` - produces site certificate and key

then the ca.crt, <site_name>.crt and <site_name>.key can be distributed to the clusterlink admin of a specific cluster.
The two steps can be automated in a central management environment.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+
+The root ca certificates will be saved in the directory ```~/.config/clustrlink/<fabric_name>```.
+
+### Install ClusterLink on K8s Cluster
+The easiest way to install ClusterLink on the K8s cluster is by using the command:
+
+    clusterlink install --peer <name>
```
elevran, line 59: Need to think further about the install step. Is it only filling in the CRD (where the operator already runs and the CRD kind already defined - but no CRD instance exists).

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+
+This command installs the ClusterLink operator and creates the ClusterLink deployment.  
+Alternatively you can do it in steps:
+First, create and deploy the peer certificate:
+
+    clusterlink create peer --name <name>
```
elevran, line 65: where would this command get the CA key and certificate needed to sign the new site certificate?
I think we might be mixing users and their access rights (e.g., site admin should not have access to a CA private key, fabric admin should not have access to a specific cluster to run `kubectl` against it...)

The steps should be much better delineated by user.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+    kubectl create -f ```~/.config/clustrlink/<fabric_name>/<peer_nam>/secret.yaml```
+
+Second, create the ClustreLink operator:
+
+    kubectl create -f config/setup/
```
elevran, line 70: we can also have a single YAML file with multiple objects

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+  namespace: clusterlink-ns
+  name: peer1
+spec:
+  dataplaneType: "envoy"
+  dataplaneReplicates: 1
+EOF
+```
+The operator will create:
+* ClusterLink controlplane deployment, including controlplane-pod, controlplane-service, and RBAC roles.  
+* ClusterLink dataplane deployment, including dataplane-pod and dataplane-service.
```
elevran, line 87: single or multiple Pods for the data plane?

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+- Define the task functions of the ClusterLink operator.
+- Define the CRD (Custom Resource Definition) for the ClusterLink operator.
+
+## Non-Goals
+The following aspects are explicitly excluded and are out of scope in the current design:
+
+- This document focuses only on the deployment of ClusterLink to the K8s cluster. 
+The deployment to another environment as VMs is out of scope. 
+- The security aspects and user permissions concerning the creation of certificates, specifically for fabric or peer.
+## Proposal
```
elevran, line 109: we should start running YAML linting. While it is displayed fine, there are expectations around use of empty lines between sections, etc.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+The ClusterLink deployment will be done in the following steps, (as described in more detail in the User section):
+
+1. Create a Fabric certificate if it does not exist.
+2. Create the ClusterLink operator.
+3. Deploy the ClusterLink CRD to the operator, which will create peer certificates, ClusterLink deployments (control plane and data plane), services, and external access point.
+
```
elevran, line 115: Please add discussion of namespaces being used.
- operator namespace
- clusterlink data/control namespace
- user namespaces

This relates to the system user discussion as well and impacts RBAC provided for each of (1) operator; (2) control plane; and (3) data plane.
You should discuss RBAC under security section below.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+- **Spec Section:**
+  
+    | Field name | Description | Default value |
+    | ---- | ----- | ------ |
+    |dataplaneType| Types of dataplane, supported values "go" or "envoy"|"envoy"|
+    |dataplaneReplicates| Number of dataplane replicas|1|
+    |logLevel| Log level severity| "info"|
+    |logFile| Path to a file where logs will be written| By default logs will be printed to stderr|
+    |image| The container registry to pull the project images. If empty, it will use the local registry| ghcr.io/clusterlink-net.
```
elevran, line 128: consider adding an `image-tag` string field to select version.
For now it is ok to assume the image name as hard-coded and also the same tag for both control and data planes. Ultimately you may want to upgrade only one (mostly during development).
Future: do we want to support automatic updates when `image-tag` is set to `latest`? Or is `latest` not a supported value?

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+### ClusterLink CRD
+The ClusterLink CRD includes the following fields:
+- Api version: clusterlink.net/v1alpha1
+- Kind: clusterlink
+
+- **Spec Section:**
+  
+    | Field name | Description | Default value |
+    | ---- | ----- | ------ |
+    |dataplaneType| Types of dataplane, supported values "go" or "envoy"|"envoy"|
```
elevran, line 125: FYI: there is a way to tell the CRD "compiler" to validate these are the only allowed values as user input.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+    |dataplaneReplicates| Number of dataplane replicas|1|
```
elevran, line 126: dataplaneReplicas? 
what do you think about splitting the spec into shared values for both data and control and then having Spec.DataPlane and Spec.ControlPlane objects for the specific deployment (e.g., we would then have `Spec.DataPlane.replicas` but `image` might be shared)

design-proposals/project-deploymnet.md
elevran, line 128: have you looked at what the [image deploy plugin](https://book.kubebuilder.io/plugins/deploy-image-plugin-v1-alpha) provides?

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+    |image| The container registry to pull the project images. If empty, it will use the local registry| ghcr.io/clusterlink-net.
+    |serviceExposureType| Type of service to expose ClusterLink deployment, supported values: "load-balancer", "API-gateway", "node-port". |load-balancer|
+
+- **Status section:**
+
+    Todo
+## Impacts / Key Questions
+
+- Do we need the ClusterLink CLI? The only command that seems necessary is for fabric and peer certificates.
+- We don't use or initialize gwctl anymore.
```
elevran, line 137: Is this a question or statement?
I don't think initialization is required since we're switching over to kubectl and it has its own configuration.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+- **Spec Section:**
+  
+    | Field name | Description | Default value |
+    | ---- | ----- | ------ |
+    |dataplaneType| Types of dataplane, supported values "go" or "envoy"|"envoy"|
+    |dataplaneReplicates| Number of dataplane replicas|1|
+    |logLevel| Log level severity| "info"|
+    |logFile| Path to a file where logs will be written| By default logs will be printed to stderr|
+    |image| The container registry to pull the project images. If empty, it will use the local registry| ghcr.io/clusterlink-net.
+    |serviceExposureType| Type of service to expose ClusterLink deployment, supported values: "load-balancer", "API-gateway", "node-port". |load-balancer|
```
elevran: `exposureType`? 
Are there values that are needed based on the type of exposure? For example, do we want a configurable service name or use a hard coded (e.g.,) "load-balancer"/"node-port"?
The other end of the exposure should point to the control plane service, correct?

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+- We don't use or initialize gwctl anymore.
+- Need to have a security discussion focusing on how and who deploys the peer and fabric certificates. 
+
+## Risks and Mitigations
+
+Not applicable.
+
+### Security Considerations
+
+The assumption in this design document is that there isn't a distinction in the security level between creating fabric certificates and peer certificates. Users with permission for one are assumed to have permission for the other.
```
elevran, line 146: I believe they should be distinct.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+    |dataplaneType| Types of dataplane, supported values "go" or "envoy"|"envoy"|
+    |dataplaneReplicates| Number of dataplane replicas|1|
+    |logLevel| Log level severity| "info"|
+    |logFile| Path to a file where logs will be written| By default logs will be printed to stderr|
+    |image| The container registry to pull the project images. If empty, it will use the local registry| ghcr.io/clusterlink-net.
+    |serviceExposureType| Type of service to expose ClusterLink deployment, supported values: "load-balancer", "API-gateway", "node-port". |load-balancer|
+
+- **Status section:**
+
+    Todo
```
elevran, line 133: Need a definition...

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+- The security aspects and user permissions concerning the creation of certificates, specifically for fabric or peer.
+## Proposal
+The ClusterLink deployment will be done in the following steps, (as described in more detail in the User section):
+
+1. Create a Fabric certificate if it does not exist.
+2. Create the ClusterLink operator.
+3. Deploy the ClusterLink CRD to the operator, which will create peer certificates, ClusterLink deployments (control plane and data plane), services, and external access point.
+
+### ClusterLink CRD
+The ClusterLink CRD includes the following fields:
```
elevran, line 117: You should discuss what the operator does in response to the CRD:
- created
- modified (can be same as created)
- deleted

You should also mention whether you're doing bidirectional watching and sync'ing or not (e.g., if a user deletes the control / data plane deployments or services - do you recreate them?). It is ok not to do it (e.g., since we expect this to be a privileged operation by a knowledgeable operator and the dedicated namespace (`clusterlink-system`) is protected. Just be explicit. Can be added as future.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+Not applicable.
+
+### Security Considerations
+
+The assumption in this design document is that there isn't a distinction in the security level between creating fabric certificates and peer certificates. Users with permission for one are assumed to have permission for the other.  
+Additionally, the Fabric certificate should be accessible to the k8s operator by passing through k8s secrets during the creation of peer certificates.
+
+## Future Milestones
+In the first step, the k8s operator will be built and will utilize the current cl-adm implementation to create peer and fabric certificates.  
+The `cl-adm create peer1` command will also generate the files: k8s-secret.yaml (containing all the certificates for the control-plane and data-plane) and clusterlink.yaml (containing the manifests for the cl-operator).
```
elevran, line 151: do you propose to keep the `cl-adm` code or executable? That is, keeping the executable with all subcommands or just the package that creates the needed certificates.

elevran (changes requested):
Thanks for working on this!

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,174 @@
...
+**Author**: @kfirtoledo
+
+**Begin Design Discussion**: 2023-11-30
+
+**Status:** draft
+
+## Summary/Abstract
+
+This design proposal outlines the deployment process for the ClusterLink project on a Kubernetes (K8s) cluster.  
+The proposal specifically addresses deployment within the K8s environment.\
```
elevran, line 12: trailing `\`?
Also not clear what this sentence adds to the previous.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,174 @@
...
+It will rely on a dedicated K8s operator, to make the deployment more user-friendly, simplified and native to K8s users.
+
+## Background
+
+### Motivation and problem space
+
+This design proposal aims to simplify and enhance the deployment of the ClusterLink project.
+By leveraging K8s native mechanisms, such as operators and CRDs, the objective is to establish a user-friendly deployment process.\
```
elevran, line 20: More trailing `\` at end of line. This seems to be repeated throughout the document and should be fixed..

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,174 @@
...
+Currently, ClusterLink deployment involves using the cl-adm CLI, which generates a large YAML file that contains the control plane and dataplane deployments, services, and RBAC configurations.\
+In the new design, this process will be managed by the ClusterLink operator, which will consume a simpler and more concise YAML configuration.
```
elevran, line 22: In addition, 
- the use of CRD and operator help in automating the deployment changes over time;
- the proposal differentiates between fabric and site administrators and provides more secure deployment model via k8s RBAC (i.e., `cl-adm` requires access to a shared file system for fabric and site administration).

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Impact and desired outcome
+
+Implementing this proposal is expected to improve the accessibility and ease of use for ClusterLink users.\
```
elevran, line 26: Unsure about the use of `accessibility` in this context

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,174 @@
...
+Aligning the deployment process with K8s standards will particularly benefit use cases involving central control management.
+
+### Prior discussion and links
+
+Not applicable.
+
+## User/User Story
+
+The purpose of ClusterLink is to interconnect various clusters within the same fabric.\
```
elevran, line 35: Start with "As a site administrator I would like to join my cluster to an existing fabric" and then add the information written as additional context?

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,174 @@
...
+The deployment of ClusterLink is done by the cluster owner.\
+Each cluster owner is required to obtain the Fabric CA certificate and the cluster certificates (private and public keys) from the fabric administrator.
+
+## Goals
+
+This design document should:
+
+* Define the steps for deploying a ClusterLink gateway to a K8s cluster.
+* Define the task functions of the ClusterLink operator.
+* Define the CRD (Custom Resource Definition) for the ClusterLink operator.
```
elevran, line 46: `... the ClusterLink *deployment* operator." ? 
Should we be clear in explicitly focusing on deployment vs runtime configuration?

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,174 @@
...
+## Proposal
+
+The ClusterLink deployment will be done in the following steps:
+
+1. Deploy the cluster certificates as k8s secrets to the cluster.
+   The certificate include the public Fabric CA certificates and the site certificate that generated by the fabric administrator.
+2. Create the ClusterLink operator.
+3. Deploy the ClusterLink CRD to the operator ClusterLink deployments that includes:
+
+   * ClusterLink controlplane deployment, including controlplane-pod, controlplane-service, and RBAC roles.
```
elevran, line 65: As worded, this suggests that these fields (e.g., RBAC roles) are part of the CRD.
Did you mean that they're created by the operator in response to the CRD?

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,174 @@
...
+
+   * ClusterLink controlplane deployment, including controlplane-pod, controlplane-service, and RBAC roles.  
+   * ClusterLink dataplane deployment, including dataplane pods, whether single or multiple, along with the dataplane-service..
+   * ClusterLink external access point using load-balancer/node-port/gateway-API service.
+  
+Overall, the ClusterLink deployment stages are:
+
+<img src="deployment.png" width="800" height="400" alt="ClusterLink Deployment Stages"/>
+
+The operator will be created in a dedicated namespace and, by default, will be ```cl-operator-ns```. It will have privileged permissions, enabling it to create ClusterLink components within the dedicated namespace.
```
elevran, line 73: "... it will have permissions to create ClusterLink components...."

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,174 @@
...
+The ClusterLink components will be created in ```clusterlink-system-ns```, and the control-plane will have privileged permissions to execute operations on the ClusterLink CRDs, such as peer, export, import, policy, etc.
+Additionally, the operator will create and update components in response to the ```clusterlink.yaml``` create or update actions. Once created or updated, the operator will not actively monitor the state of each component, assuming that any changes will be made only by privileged users.
+Furthermore, the operator will delete all components and the namespace of ClusterLink when the ```clusterlink.yaml``` file is deleted.
```
elevran, line 76: This is misleading.
The `clusterlink.yaml` contains an instance of the deployment CRD. It is unrelated to the file system.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,174 @@
...
+
+The deployment file for the operator and example of ```clusterlink.yaml``` will be included in every project release.
+
+The ClusterLink CLI will be utilized to automate the deployment process.\
+For instance, the following commands can assist the fabric administrator in creating certificates in the directory ```~/.config/clustrlink/<fabric_name>```:
```
elevran, line 81: - any reason not to generate these in the local directory? Or to allow `--output` CLI option?

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,174 @@
...
+
+    clusterlink create fabric --name <fabric_name>
+    clusterlink create site --name <site_name> --fabric <fabric_name> 
+
+Automation for deploying the certificates as a secret to the cluster, deploying the ClusterLink operator, and the ```clusterlink.yaml``` can be achieved with
+
+    clusterlink install --site <site_name>
```
elevran, line 88: How will you make the cluster kubeconfig accessible to the CLI?
What is the meaning of `site` here?

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,174 @@
...
+
+### ClusterLink CRD
+
+The ClusterLink CRD includes the following fields:
+
+* Api version: clusterlink.net/v1alpha1
```
elevran, line 94: API

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,174 @@
...
+* Kind: clusterlink
+
+* **Spec Section:**
+  
+    | Field name | Description | Default value |
+    | ---- | ----- | ------ |
+    |dataplane.type| Types of dataplane, supported values "go" or "envoy"|"envoy"|
```
elevran, line 101: I don't think the use of `.` in field names is common. Unless you meant that there's a sub-object?

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,174 @@
...
+    |dataplane.replicates| Number of dataplane replicas|1|
+    |logLevel| Log level severity| "info"|
+    |containerRegistry| The container registry to pull the project images. If empty, it will use the local registry| ghcr.io/clusterlink-net|
```
elevran, line 104: if there's a default, it can't be empty to denote local registry

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,174 @@
...
+    |imageTag| The project images version | latest|
+    |serviceType| Type of service to expose ClusterLink deployment, supported values: "load-balancer", "API-gateway", "node-port". |load-balancer|
+
+* **Status section:**
+
+    | Field name | Description | Default value |
+    | ----------- | ----------- | ------------- |
+    | status      | Status of the components controlled by the operator. Supported values: <br> None - if all components have not been created. <br> Ready - if all components are in a ready and running status. <br> Error - If one of the components is in an error status. <br> Update - if one of the components is in an update status. | None |
```
elevran, line 112: maybe have conditions for each component?

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,174 @@
...
+    | message     | Optional message that provides additional information about the status. | None |
+
+Example to clusterlink CRD:
+```
+apiVersion: clusterlink.net/v1alpha1
+kind: ClusterLink
+metadata:
+  namespace: clusterlink-ns
```
elevran, line 120: Is the CRD deployed into the operator or clusterlink namespace?

design-proposals/project-deploymnet.md
elevran: It may be confusing to keep `cl-adm`

elevran (changes requested):

design-proposals/project-deploymnet.md
kfirtoledo: The default is ghcr.io/clusterlink-net

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,169 @@
...
+Implementing this proposal is expected to improve and simplify the deployment process for the ClusterLink users.
+Aligning the deployment process with K8s standards will particularly benefit use cases involving central control management.
+
+### Prior discussion and links
+
+Not applicable.
+
+## User/User Story
+
+As a site administrator, I would like to join my cluster to an existing fabric. The fabric administrator should send me the certificates required to join the fabric: the fabric CA certificate and the cluster certificates (private and public keys). Using these certificates and the ClusterLink operator, I can deploy the ClusterLink system in my cluster. Now, any user in my cluster can connect to other remote services in the fabric, even if they are located on different locations, or different cloud providers.
```
elevran, line 39: Another potential story is deploy / undeploy of ClusterLink. It somewhat overlaps with the above, but is more focused on what is being addressed by this design (IMHO, the "joining a fabric" is more about the exchange of certificates and *then* the deployment - but maybe that's just my world view...)

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,169 @@
...
+
+The following aspects are explicitly excluded and are out of scope in the current design:
+
+* This document focuses only on the deployment of ClusterLink to the K8s cluster.
+The deployment to another environment as VMs is out of scope.
+* Security considerations related to the creation of certificates for the fabric or each site by the fabric administrator are also beyond the scope of this document.
+
+## Proposal
+
+The proposal describes how the ClusterLink project can be deployed on any K8s Cluster. In this proposal, we distinguish between two entities: first, the fabric administrator, who manages the fabric and is responsible for determining which clusters can join the fabric, and second, the site administrator, who manages the k8s Cluster and is responsible for deploying the ClusterLink into the cluster.
```
elevran, line 59: nit: may be easier to grok if you split the two roles into different bullets instead of a long, multipart, sentence...

design-proposals/project-deploymnet.md
elevran, line 39: Another potential story is deploy / undeploy of ClusterLink. It somewhat overlaps with the above, but is more focused on what is being addressed by this design (IMHO, the "joining a fabric" is more about the exchange of certificates and then the deployment - but maybe that's just my world view...)

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,169 @@
...
+
+Before deploying the ClusterLink, a few prerequisite steps should be completed:
+
+1. The fabric administrator should create the site certificates (public and private) and transfer them, along with the public CA certificate, to the site administrator.
+1. The site administrator should deploy the certificates as k8s secrets to the cluster.
+1. The site administrator should deploy the ClusterLink operator to the cluster.
+
+After completing all the prerequisites, the site administrator can deploy the ClusterLink Custom Resource Definition (CRD) to the operator. The operator will then create the ClusterLink components, including:
```
elevran, line 67: There are two separate objects in play:
- the CRD (i.e., the class)
- a CRD object (i.e., the instance).

The class can/should be registered with the API server before the operator is deployed (since the operator will attempt to create a watch on the object, the object schema - or CRD - must already be defined). Both should be done before the site administrator can create the CRD instance that defines the desired deployment configuration)

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,169 @@
...
+
+* A ClusterLink dedicated namespace, which by default will be cl-operator-ns.
```
elevran, line 69: what happens when >1 CRD instance is created? How do we handle it (or not)?

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,169 @@
...
+* Deployment of ClusterLink controlplane components, including controlplane-pod, controlplane-service, and RBAC roles.
+* Deployment of ClusterLink dataplane components, including dataplane pods (single or multiple) and the dataplane-service.
+* Deployment of ClusterLink ingress, providing an external access point using load-balancer/node-port/gateway-API service."
+
+Overall, the ClusterLink deployment stages are:
+
+<img src="deployment.png" width="800" height="400" alt="ClusterLink Deployment Stages"/>
+
+The ClusterLink operator will have privileged permissions, enabling it to create ClusterLink components within the dedicated namespace.
```
elevran, line 78: It will also have namespace creation privilege so it can place all components into the namespace.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,169 @@
...
+The ClusterLink components will be created in `clusterlink-system-ns`, and the control-plane will have privileged permissions to execute operations on the ClusterLink CRDs, such as peer, export, import, policy, etc.
```
elevran, line 79: The other CRDs (such as `Peer`, `Import`, and `Policy`) are not in scope of this design and should not be mentioned, IMHO.
We will likely decide to have a separate "runtime configuration" operator as part of the control plane, so the deployment operator is separate and need not access these.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,169 @@
...
+Additionally, the operator will create and update components in response to the `clusterlink-system.yaml` create or update actions. Once created or updated, the operator will not actively monitor the state of each component, assuming that any changes will be made only by privileged users.
```
elevran, line 80: Please refrain from mentioning YAMLs. The operator watches and operates on the CRD instance - and does not care if/how it was created (manually, by `kubectl apply` file or by `kubectl edit`).

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,169 @@
...
+Furthermore, the operator will delete all components and the namespace of ClusterLink when the `clusterlink-system.yaml` file is deleted.
```
elevran, line 81: ditto - "... when the CRD instance is deleted" - it is not aware of any file being involved in the process.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,169 @@
...
+
+The deployment file for the operator and example of `clusterlink-system.yaml` will be included in every project release.
+
+The ClusterLink CLI will be utilized to automate the deployment process.
```
elevran, line 85:
```suggestion
The ClusterLink CLI can be used to automate some of the the deployment process steps.
```

design-proposals/project-deploymnet.md
elevran: So be explicit about that. Or at lease pause to get user acceptance when deploying to the last used kubeconfig. Another option is to explicitly require the kubeconfig and context as parameters.

elevran (changes requested):
I think it's almost ready for approval. 
Main points I would address is the reference to YAML files and the removal of runtime configuration CRDs as needed/accessible to this operator.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+* Define the steps for deploying a ClusterLink gateway to a K8s cluster.
+* Define the task functions of the ClusterLink deployment operator for setting up the ClusterLink components in the cluster.
+* Define the CRD (Custom Resource Definition) for the ClusterLink deployment operator.
+
+## Non-Goals
+
+The following aspects are explicitly excluded and are out of scope in the current design:
+
+* This document focuses only on the deployment of ClusterLink to the K8s cluster.
+The deployment to another environment as VMs is out of scope.
```
elevran, line 54:
```suggestion
The deployment to another environment, such as VMs, is out of scope.
```

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+Implementing this proposal is expected to improve and simplify the deployment process for the ClusterLink users.
+Aligning the deployment process with K8s standards will particularly benefit use cases involving central control management.
+
+### Prior discussion and links
+
+Not applicable.
+
+## User/User Story
+
+As a site administrator, I would like to join my cluster to an existing fabric. The fabric administrator should send me the certificates required to join the fabric: the fabric CA certificate and the cluster certificates (private and public keys). Using these certificates and the ClusterLink operator, I can deploy the ClusterLink system in my cluster. Now, any user in my cluster can connect to other remote services in the fabric, even if they are located on different locations, or different cloud providers.
```
elevran, line 39:
```suggestion
The below user stories describe sample needs of a site administrator. Fabric administrators' requirements aren't
 elaborated.
 
1. As a site administrator, I would like to join my cluster to an existing fabric. After receiving the credentials (e.g.,
 fabric certificate and site keypair, I would like to provide minimal additional configuration needed to deploy
 ClusterLink to the site/cluster I manage.
1. As a site administrator, I would like to scale my ClusterLink deployment to meet growing demand (e.g., bandwidth)
 for cross-site communication.
1. As a site administrator, I would like to completely remove ClusterLink from my cluster, restoring its previous state.
```

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+In this proposal, we distinguish between two entities:
+
+1. Fabric administrator: manages the fabric and is responsible for determining which clusters can join the fabric.
+1. Site administrator: manages the k8s Cluster and is responsible for deploying the ClusterLink into the cluster.
+
+Before deploying the ClusterLink, a few prerequisite steps should be completed:
+
+1. The fabric administrator should create the site certificates (public and private) and transfer them, along with the public CA certificate, to the site administrator.
+2. The site administrator should deploy the certificates as k8s secrets to the cluster.
+3. The site administrator should deploy the ClusterLink operator to the cluster and register ClusterLink Custom Resource Definition (CRD) class to the cluster.
```
elevran, line 69:
```suggestion
3. The site administrator should deploy the ClusterLink operator to the cluster and register ClusterLink Custom Resource Definition (CRD) class to the cluster. This step is not dependent on the previous two and can be at the administrator's convenience.
```

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+
+After completing all the prerequisites, the site administrator can deploy the ClusterLink CRD instance to the operator. The operator will then create the ClusterLink components, including:
```
elevran, line 71:
```suggestion
After completing all the prerequisites, the site administrator can edit and apply a ClusterLink CRD instance to the operator. The operator will then create the ClusterLink components, including:
```

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+The proposal describes how the ClusterLink project can be deployed on any K8s Cluster.
+In this proposal, we distinguish between two entities:
+
+1. Fabric administrator: manages the fabric and is responsible for determining which clusters can join the fabric.
+1. Site administrator: manages the k8s Cluster and is responsible for deploying the ClusterLink into the cluster.
+
+Before deploying the ClusterLink, a few prerequisite steps should be completed:
+
+1. The fabric administrator should create the site certificates (public and private) and transfer them, along with the public CA certificate, to the site administrator.
+2. The site administrator should deploy the certificates as k8s secrets to the cluster.
```
elevran, line 68:
```suggestion
2. The site administrator should deploy the certificates as k8s secrets to the cluster. This step can be automated to reduce operator toll and chance of introducing errors.
```

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+3. The site administrator should deploy the ClusterLink operator to the cluster and register ClusterLink Custom Resource Definition (CRD) class to the cluster.
+
+After completing all the prerequisites, the site administrator can deploy the ClusterLink CRD instance to the operator. The operator will then create the ClusterLink components, including:
+
+* A ClusterLink dedicated namespace, which by default will be cl-operator-ns.
+* Deployment of ClusterLink controlplane components, including controlplane-pod, controlplane-service, and RBAC roles.
+* Deployment of ClusterLink dataplane components, including dataplane pods (single or multiple) and the dataplane-service.
+* Deployment of ClusterLink ingress, providing an external access point using load-balancer/node-port/gateway-API service."
```
elevran, line 76:
```suggestion
* Deployment of ClusterLink ingress, providing an external access point using load-balancer/node-port/gateway-API service.
```

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+
+Overall, the ClusterLink deployment stages are:
+
+<img src="deployment.png" width="800" height="400" alt="ClusterLink Deployment Stages"/>
+
+The ClusterLink operator will have privileged permissions, allowing it to create a dedicated namespace and the ClusterLink components within.
+The ClusterLink components will be created in `clusterlink-system-ns`, and the control-plane will have privileged permissions within the namespace.
```
elevran, line 83: Actually the control plane needs to watch objects in other namespaces as well. 
Also, why does it need to have privileges on its own namespace?

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+
+The ClusterLink CLI will be utilized to automate the deployment process.
+For instance, the following commands can assist the fabric administrator in creating the site certificates:
+
+    clusterlink create fabric --name <fabric_name>
+    clusterlink create site --name <site_name> --fabric <fabric_name> 
+
+For the site administrator, automation of deploying the certificates as a secret to the cluster, deploying the ClusterLink operator, and the CRD instance can be achieved with:
+
+    clusterlink install --site <site_name>
```
elevran, line 98: This is a good point to provide the certificates and keys needed as input and deploy them as secrets. You should consider if these are provided by the CRD or as parameters here.
Since the CRD does not reference secret objects, I assume their names are fixed. Thus we would want to automate their creation so the user is not assigning different names by mistake.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+Furthermore, the operator will delete all components and the namespace of ClusterLink when the CRD instance is deleted.
+The ClusterLink deployment is limited to one instance per cluster. In the case of deploying more than one CRD instance, only the first one will be taken into account, and the others will be ignored.
+
+The deployment file for the operator and example of CRD instance will be included in every project release.
+
+The ClusterLink CLI will be utilized to automate the deployment process.
+For instance, the following commands can assist the fabric administrator in creating the site certificates:
+
+    clusterlink create fabric --name <fabric_name>
+    clusterlink create site --name <site_name> --fabric <fabric_name>
```
elevran, line 94: this CLI should probably access a CA key pair so it can sign the site certificate. Unless you're assuming something about the path / names of the files based on the `fabric_name`. You may also want to have specific names for the output files as additional parameters (or, e.g., fixed to `site_name`.crt and `site_name`.key - in which case you should warn if the files exist and you're overriding them. You should probably warn on overwrite always...)

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+* **API version:** clusterlink.net/v1alpha1
+* **Kind:** clusterlink
+
+* **Spec:**
+  
+    | Field name | Subfield name| Description | Default value |
+    | ---- | ----- | ------ | ----|
+    |dataplane | | ClusterLink dataplane component attributes||
+    |  |type| Types of dataplane, supported values "go" or "envoy"|"envoy"|
+    | |replicates| Number of dataplane replicas|1|
```
elevran, line 115:
```suggestion
    | |replicas| Number of dataplane replicas|1|
```
Or `count`? See what ReplicaSet uses?

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+    |ingress|| ClusterLink ingress component attributes ||
+    ||type| Type of service to expose ClusterLink deployment, supported values: "load-balancer", "API-gateway",    "node-port". |load-balancer|
```
elevran, line 117: I would stick to the k8s spelling and capitalization of these. For example, `LoadBalancer`, `NodePort`, `Gateway`, `Ingress`, ...

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+    |logLevel| |Log level severity for all the components (controlplane and dataplane)| "info"|
+    |containerRegistry| |The container registry to pull the project images when the images is not present locally | ghcr.io/clusterlink-net|
+    |imageTag| |The project images version | latest|
+
+* **Status:**
+
+    | Field name | Description | Default value |
+    | ----------- | ----------- | ------------- |
+    | controlplaneStatus      | Status of the controlplane components controlled by the operator. Supported values: <br> None - if all components have not been created. <br> Ready - if all components are in a ready and running status. <br> Error - If one of the components is in an error status. <br> Update - if one of the components is in an update status. | None |
```
elevran, line 126: - No need for XYZStatus field name as it is implied by being under the `Status` object
- Consider using k8s Conditions

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,176 @@
...
+
+Not applicable.
+
+### Security Considerations
+
+For the deployment of ClusterLink, the Fabric CA certificate (public key) and the peer certificates (public and private keys) must be provided. These certificates will be deployed to the cluster using k8s-secrets and should be handled by the cluster owner. The creation of these certificates can be performed by the central management, which is outside the scope of this document.
+
+The ClusterLink operator will have privileged permissions to deploy the ClusterLink project to the cluster. The control plane of ClusterLink will be granted privileged access to all cluster resources through RBAC (Role-Based Access Control).
+
+In the future, we may introduce a Namespace deployment, which will have permissions limited to a single namespace.
```
elevran, line 158:
```suggestion
In the future, we may introduce a per-Namespace deployment, which will have permissions limited to a single namespace.
```

elevran (changes requested):

elevran:
@kfirtoledo - can you please integrate all changes suggested/requested and move this to approval?

design-proposals/project-deploymnet.md
kfirtoledo: The Clusterlink automation will be the next thing after the operator.

kfirtoledo:
@elevran PTAL

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,188 @@
+# Design Proposal: Project Deployment with Kubernetes Operator
+
+**Author**: @kfirtoledo
+
+**Begin Design Discussion**: 2023-11-30
+
+**Status:** draft
```
elevran, line 7: change to `approved` once merged

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,188 @@
...
+## Background
+
+### Motivation and problem space
+
+This design proposal aims to simplify and enhance the deployment of the ClusterLink project.
+By leveraging K8s native mechanisms, such as operators and CRDs, the objective is to establish a user-friendly deployment process.
+Currently, ClusterLink deployment involves using the cl-adm CLI, which generates a large YAML file that contains the control plane and dataplane deployments, services, and RBAC configurations.
+In the new design, this process will be managed by the ClusterLink operator, which has the following advantages:
+
+* It consumes a simpler and more concise YAML configuration.
```
elevran, line 24:
```suggestion
* It consumes a simpler and more concise configuration in the form of a CRD.
```

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Prior discussion and links
+
+Not applicable.
+
+## User/User Story
+
+The below user stories describe sample needs of a site administrator. Fabric administrators' requirements aren't elaborated.
+
+1. As a site administrator, I want to join my cluster with an existing fabric. After receiving the credentials (e.g., fabric certificate and site keypair, I would like to provide minimal additional configuration needed to deploy ClusterLink to the site/cluster I manage.
```
elevran, line 41:
```suggestion
1. As a site administrator, I want to join my cluster with an existing fabric. After receiving the credentials (e.g., fabric certificate and site keypair), I would like to provide minimal additional configuration needed to deploy ClusterLink to the site/cluster I manage.
```

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,188 @@
...
+* Define the steps for deploying a ClusterLink gateway to a K8s cluster.
+* Define the task functions of the ClusterLink deployment operator for setting up the ClusterLink components in the cluster.
+* Define the CRD (Custom Resource Definition) for the ClusterLink deployment operator.
+
+## Non-Goals
+
+The following aspects are explicitly excluded and are out of scope in the current design:
+
+* This document focuses only on the deployment of ClusterLink to the K8s cluster.
+The deployment to another environment, such as VMs, is out of scope.
```
elevran, line 58:
```suggestion
 The deployment to another environment, such as VMs, is out of scope.
```
Should be combined with previous line or added in a separate line with indentation.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,188 @@
...
+
+The proposal describes how the ClusterLink project can be deployed on any K8s Cluster.
+In this proposal, we distinguish between two entities:
+
+1. Fabric administrator: manages the fabric and is responsible for determining which clusters can join the fabric.
+1. Site administrator: manages the k8s Cluster and is responsible for deploying the ClusterLink into the cluster.
+
+Before deploying the ClusterLink, a few prerequisite steps should be completed:
+
+1. The fabric administrator should create the site certificates (public and private) and transfer them, along with the public CA certificate, to the site administrator.
```
elevran, line 71:
```suggestion
1. The fabric administrator should create the site certificate (public and private keys) and transfer them, along with the public CA certificate, to the site administrator.
```

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,188 @@
...
+2. The site administrator should deploy the certificates as k8s secrets to the cluster.  This step can be automated to reduce operator toll and chance of introducing errors.
+3. The site administrator should deploy the ClusterLink operator to the cluster and register ClusterLink Custom Resource Definition (CRD) class to the cluster. This step is not dependent on the previous two and can be done at the administrator's convenience.
+
+After completing all the prerequisites, the site administrator can edit and apply a ClusterLink CRD instance to the operator. The operator will then create the ClusterLink components, including:
+
+* A ClusterLink dedicated namespace, which by default will be cl-operator-ns.
```
elevran, line 77: The operator namespace is required for running the operator itself and thus can't be something the operator creates.
Did you mean the namespace dedicated for ClusterLink components (e.g., `clusterlink-system`)?
If so, we may also want to require that it is created ahead of time and assume that CL always deploys to an existing namespace (this would also reduce the RBAC permissions needed by the operator). If we decide to accept that the namespace does not exist, we should create and also remove it (probably needs a `Finalizer` or at least a label for us to know when we create a namespace and when it pre-existed.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,188 @@
...
+* Deployment of ClusterLink controlplane components, including controlplane-pod, controlplane-service, and RBAC roles.
+* Deployment of ClusterLink dataplane components, including dataplane pods (single or multiple) and the dataplane-service.
+* Deployment of ClusterLink ingress, providing an external access point using load-balancer/node-port/gateway-API service.
+
+Overall, the ClusterLink deployment stages are:
+
+<img src="deployment.png" width="800" height="400" alt="ClusterLink Deployment Stages"/>
+
+The ClusterLink operator will have privileged permissions, allowing it to create a dedicated namespace and the ClusterLink components within.
+The ClusterLink components will be created in the `clusterlink-system-ns``, and the control-plane will have privileged permissions within the namespace (for creating K8s services) and watch permissions in other namespaces.
```
elevran, line 87:
```suggestion
The ClusterLink components will be created in the `clusterlink-system` namespace, and the control-plane will have privileged permissions within the namespace (for creating K8s services) and watch permissions in other namespaces.
```

Also, I think the control plane needs to watch for CRDs and create Services in other namespaces, not only the local namespace. It is not clear the way it's worded.

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,188 @@
...
+Additionally, the operator will create and update components in response to the CRD instance create or update actions. Once created or updated, the operator will not actively monitor the state of each component, assuming that any changes will be made only by privileged users.
```
elevran, line 88:
```suggestion
Additionally, the operator will create and update components in response to the CRD instance changes. Once created or updated, the operator will not actively monitor the state of each component (i.e., it will do a one-time reconciliation), assuming that any changes will be made only by privileged users.
```

Is this still true?

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,188 @@
...
+Furthermore, the operator will delete all components and the namespace of ClusterLink when the CRD instance is deleted.
+The ClusterLink deployment is limited to one instance per cluster. In the case of deploying more than one CRD instance, only the first one will be taken into account, and the others will be ignored.
```
elevran, line 90:
```suggestion
The ClusterLink deployment is limited to one instance per cluster. In the case of deploying more than one CRD instance, only the first one will be taken into account, and the others will be ignored and their Status shall be updated to reflect that.
```

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,188 @@
...
+* **Status:**
+
+    | Field name | Subfield name|  Description |
+    | ----------- | ----------- |----------- |
+    | controlplane || Status of the controlplane components controlled by the operator |
+    |  |Conditions|The controlplane will have two status conditions: 1. ```DeploymentReady``` will be set to true when the controlplane deployment is ready, and 2. `ServiceReady` will be set to true when the controlplane service is ready|
+    | datalplane || Status of the dataplane components controlled by the operator |
+    |  |Conditions|The dataplane will have two status conditions: 1. ```DeploymentReady``` will be set to true when the dataplane deployment is ready, and 2. `ServiceReady` will be set to true when the dataplane service is ready|
+    | ingress || Status of the dataplane components controlled by the operator |
+    |  |Conditions|The ingress will have one status conditions: `ServiceReady` will be set to true when the external ingress service is ready|
```
elevran, line 138: You may want to consider two more conditions (along with an observed generation value):
- `Reconciling`: set to True when the controller is actively working; and
- `Stalled`: when some error condition has occurred (e.g., failed call to API master, secondary CRD instances being ignored, etc.)

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,188 @@
...
+    |  |type| Types of dataplane, supported values "go" or "envoy"|"envoy"|
+    | |replicas| Number of dataplane replicas|1|
+    |ingress|| ClusterLink ingress component attributes ||
+    ||type| Type of service to expose ClusterLink deployment, supported values: "LoadBalancer", "Gateway", "NodePort", "None". |None|
+    ||Port| Port represents the port number of the external service |443 for all types,except for NodePort, where the port number will be allocated by Kubernetes | 
+    |logLevel| |Log level severity for all the components (controlplane and dataplane)| "info"|
+    |containerRegistry| |The container registry to pull the project images when the images is not present locally | ghcr.io/clusterlink-net|
+    |imageTag| |The project images version | latest|
+    |Namespace| | The namespace where the components of the ClusterLink project are deployed | clusterlink-system|
+  .
```
elevran, line 127:
```suggestion
```

design-proposals/project-deploymnet.md
```diff
@@ -0,0 +1,188 @@
...
+* **API version:** clusterlink.net/v1alpha1
+* **Kind:** clusterlink
+
+* **Spec:**
+  
+    | Field name | Subfield name| Description | Default value |
+    | ---- | ----- | ------ | ----|
+    |dataplane | | ClusterLink dataplane component attributes||
+    |  |type| Types of dataplane, supported values "go" or "envoy"|"envoy"|
+    | |replicas| Number of dataplane replicas|1|
```
elevran, line 119: Since there's no secret object reference, you may want to add to the prerequisites that the site certificate must be added  as a predefined name. You should also discuss the namespace (operator or clusterlink) since volume mounts work in the same namespace only.

elevran (approved):
Please address the open questions regarding namespace and site certificate secret and then can be merged.

PR #228, opened 2023-12-11
update website devcontainer to use features

elevran:
Add website/docs devcontainer configuration

- Container is created using "features" for installing tools
- Selection of devcontainer options will be added so vscode can switch between `dev` and `docs` which have different images


elevran:
@orozery - please take a look if you have some time. Tested on vscode, need to add an option to run the devcontainer directly from the command line as an isolated environment. This will be a separate PR

michalmalka (approved):

PR #237, opened 2023-12-19
design-proposals: CRD based managment

orozery:


orozery:
@elevran @kfirtoledo Please see added user stories. Next will write the CRDs and controllers design.

elevran:
@orozery - should #215 be closed in favor of this? seems to be a continuation of it.
I would suggest adding a `wip` label on this, regardless.

design-proposals/crd-based-management.md
```diff
@@ -0,0 +1,281 @@
...
+- The target service does not exists.
+- The requested port is not a valid port in the requested service.
+
+Like in the single-port case above, the port (copied from the spec) is reported as part of the export status.
+If any changes are made to the target service, the export status is expected to update accordingly.
+
+### Export endpoint user story
+
+User wants to export some service which is accessible from within a cluster via an endpoint host/ip:port.
+In particular, this can be used to export a service from any namespace (e.g. my-service.my-namespace.svc.cluster.local).
```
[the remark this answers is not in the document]
elevran: Or maybe we can just exclude that in the first iteration and add support later on. First iteration can be limited to same namespace services.
A workaround would be to have the user create an Service with `Spec.Type; ExternalName` in their namespace.

design-proposals/crd-based-management.md
```diff
@@ -0,0 +1,281 @@
...
+The status of the peer object will report whether the peer is currently reachable,
+and the last time it was successfully reached (via heartbeats).
+
+### Import service user story
+
+User of site B wants to import an exported service "foo" from namespace X of site A,
+into namespace Y of site B.
+
+To do so, it creates an Import k8s object in the desired namespace Y.
+The spec of the import includes one or more exports which serve as the source of the import.
```
elevran, line 121: So we forgo the notion of binding and instead can have multiple Exports? How are exports defined for an import?
Is it (e.g.) an array of strings containing Peer and export names? Object references? Something else?

design-proposals/crd-based-management.md
```diff
@@ -0,0 +1,281 @@
...
+
+TODO The reverse of some of the above...
+
+### Prior discussion and links
+
+TODO Some initial ideas in GH issue. Reference.
+
+## (Optional) User/User Story
+
+### Users
```
elevran, line 53: Maybe it is worthwhile to be explicit in defining the management objects (CRD) that are proposed and what they are defined. Or, using the current concepts, refer to an existing doc. After an initial read, I do not think you propose to carry over all current management objects from `gwctl`, but I could be wrong.
Consider having a separate section for "runtime configuration concepts", that would include such things as Peer, Export, Policy, ...

elevran (changes requested):

design-proposals/crd-based-management.md
```diff
@@ -0,0 +1,281 @@
...
+Multiple users
+
+- Fabric admin: Control over entire fabric (e.g, create fabric, add peers to fabric);
+ mostly not in scope for this design doc
+- Site admin: Control a single cluster. Deploys CL pods to the cluster to a specific namespace. Add existing peers. Add global policies (e.g. allow only a certain namespace to access a specific peer). Has full control of the kubernetes cluster (all namespaces).
+- Local importer/exporter (aka user): Creates import, exports, and policies for them. In most cases, RBAC will limit this user to work in a single unique namespace. We will support users from multiple namespaces.
+
+### Export local service user story
+
+User wants to export a single-port local service "foo" in namespace X.
```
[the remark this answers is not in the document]
orozery: In the the definition above I tried to link the third type to "user" (`Local importer/exporter (aka user)`).
I'm good with replacing "user" with a more explicit type, but first let's agree on how we call this importer/exporter user.

design-proposals/crd-based-management.md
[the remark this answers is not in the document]
orozery: > Or maybe we can just exclude that in the first iteration and add support later on. First iteration can be limited to same namespace services. A workaround would be to have the user create an Service with `Spec.Type; ExternalName` in their namespace.

Agreed. And this actually shows that you cannot enforce access restriction to services, as the user can create a local service which redirects to another service from another namespace.

design-proposals/crd-based-management.md
```diff
@@ -0,0 +1,281 @@
...
+
+### Peer creation user story
+
+Site A admin wants to allow users to import services from site B.
+
+To do so, it creates a Peer k8s object for site B.
+The name of the object will be used by importers as part of the routable name of an export.
+The peer spec includes one or more gateways (host/ip,port) for accessing the peer's dataplane server.
+
+**QUESTION**: Should we add an expected server name to the spec, or rely just on the gateway hosts?
```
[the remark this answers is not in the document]
orozery: Yes, this is the name you expect to see on the peer's certificate.

design-proposals/crd-based-management.md
orozery: See the next sentence:
`Each export is identified using its routable name: (peer name, export namespace, export name).`

design-proposals/crd-based-management.md
```diff
@@ -0,0 +1,281 @@
...
+and the last time it was successfully reached (via heartbeats).
+
+### Import service user story
+
+User of site B wants to import an exported service "foo" from namespace X of site A,
+into namespace Y of site B.
+
+To do so, it creates an Import k8s object in the desired namespace Y.
+The spec of the import includes one or more exports which serve as the source of the import.
+Each export is identified using its routable name: (peer name, export namespace, export name).
```
kfirtoledo, line 122: What happens:
1. If he wants to export from multiple peers? need 2 imports?
2. If he doesn't provide the namespace?
3. If we have multiple ports? Do we open a connection per port?

design-proposals/crd-based-management.md
michalmalka, line 1: Should be Create (and not Created)

elevran:
Content and discussion copied over into original issue #215

PR #244, opened 2024-01-03
Set search placeholder (for en)

elevran:
Replace the default search placeholder (`Search this site...`) with an empty string so only magnifying glass shows


michalmalka (approved):
Nice and Simple

PR #247, opened 2024-01-04
Enable tagliatelle linter

elevran:
- changes to config struct are internal and not exposed to user anyway
- added linter ignores where it made sense (e.g., follow k8s convention...)

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -60,7 +60,7 @@ linters:
     - revive
     - staticcheck
     - stylecheck
-    # - tagliatelle
+    - tagliatelle
     - tenv
     - testableexamples
     # - thelper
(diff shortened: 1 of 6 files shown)


zivnevo:
What is the motivation to convert json keys to snake case?
I believe camel case is also pretty common. See for example [Google's JSON style guide](https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format).
It's also the linter's default (which we currently override in `.golangci.yaml:115`)

elevran:
I don't have a preference for camelCase or snake_case, the only motivation is consistency across all files (which we currently do not have).
Will change to use `tagliatelle` default naming conventions (e.g., camelCase for JSON) and address accordingly

elevran:
@zivnevo PTAL

zivnevo (approved):
LGTM

PR #250, opened 2024-01-04
enable nestif linter

elevran:
- uncomment linter
- simplify nested if/else to align happy path to left where possible

Note to reviewer: **please double check that I did not change the logic!**


kfirtoledo (approved):

pkg/platform/k8s/reconciler.go
```diff
@@ -54,31 +54,30 @@ func (r *Reconciler) CreateResource(name string, obj client.Object) {
...
+	if !ok {
+		r.logger.Errorf("error occurred while updating K8s %v: %v is not exist", reflect.TypeOf(obj).String(), name)
+	}
+
+	if obj.GetName() == oldObj.GetName() {
+		err := r.client.Update(context.Background(), obj)
+		if err != nil {
+			r.logger.Errorf("error occurred while updating K8s %v %v:", reflect.TypeOf(obj).String(), err)
+			r.failedList[name] = reconcileObj{spec: obj, op: updateOp}
 			return
```
orozery, line 67: Why did you move the return inside the error case?

elevran: should processing (create resource) continue when the Update failed? Does it depend on the error returned?
Perhaps I misunderstood the logic... Let me try to revert this portion by removing the return and see if tests pass

PR #259, opened 2024-01-10
api: Introduce Clusterlink operator API.

kfirtoledo:
Introduce the operator API that will be used by the ClusterLink operator.
Note: The zz_generated.deepcopy.go file is an autogenerated file from the API, used for copying the API Go object.


api/v1alpha1/operator_types.go
```diff
@@ -0,0 +1,122 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   OperatorSpec   `json:"spec,omitempty"`
+	Status OperatorStatus `json:"status,omitempty"`
+}
+
+//+kubebuilder:object:root=true
+
+// OperatorList contains a list of Operator.
```
orozery, line 113:
```suggestion
// OperatorList represents a list of Operator objects.
```

api/v1alpha1/operator_types.go
```diff
@@ -0,0 +1,122 @@
...
+type OperatorList struct {
```
orozery, line 114: Do we really need this struct?

api/v1alpha1/operator_types.go
```diff
@@ -0,0 +1,122 @@
...
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []Operator `json:"items"`
+}
+
+func init() {
+	SchemeBuilder.Register(&Operator{}, &OperatorList{})
+}
```
orozery, line 122: The other projects I've seen use this in `register.go` (i.e. `groupversion_info.go`) instead:
```
var (
	// SchemeBuilder points to a list of functions added to Scheme.
	SchemeBuilder      runtime.SchemeBuilder
	localSchemeBuilder = &SchemeBuilder
	// AddToScheme applies all the stored functions to the scheme.
	AddToScheme = localSchemeBuilder.AddToScheme
)

func init() {
	localSchemeBuilder.Register(addKnownTypes)
}

// Adds the list of known types to api.Scheme.
func addKnownTypes(scheme *runtime.Scheme) error {
	scheme.AddKnownTypes(SchemeGroupVersion,
		&Operator{},
	)
	metav1.AddToGroupVersion(scheme, SchemeGroupVersion)
	return nil
}
```

orozery (changes requested):
@kfirtoledo I see in other big projects that they use `pkg/apis/<group>/<version>` instead of this PR `pkg/api/<version>`.
Also, they use `register.go` instead of `groupversion_info.go`.

kfirtoledo:
> @kfirtoledo I see in other big projects that they use `pkg/apis/<group>/<version>` instead of this PR `pkg/api/<version>`. Also, they use `register.go` instead of `groupversion_info.go`.

1. I put it under `api` and not `pkg/api`- Personally I don't have any preference.
2. About the group I used clustrelink.net and I thought that all the CRDs would be under that group so we don't need a folder for that.
3. `groupversion_info.go`- was generated by the kubebuilder, so we need to consider changing its name.

orozery:
> 2. About the group I used clustrelink.net and I thought that all the CRDs would be under that group so we don't need a folder for that.

Perhaps it's good to use `pkg/apis/<group>` even though we have a single group, just so that if in the future we'll add a new group, we won't have to break external packages who may import this.

api/v1alpha1/operator_types.go
```diff
@@ -0,0 +1,122 @@
...
+	Type string `json:"type,omitempty"`
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=10
+	// +kubebuilder:default=1
+	// Replicas represents the number of dataplane replicas.
+	Replicas int `json:"replicas,omitempty"`
+}
+
+// IngressSpec defines the type of the ingress component in ClusterLink.
+type IngressSpec struct {
```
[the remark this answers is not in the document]
kfirtoledo: No, for now. But I like the separation into 3 components: control plane,dataplane, and ingress.

api/v1alpha1/operator_types.go
kfirtoledo: This was generated by the kubebuilder. But it makes sense because it is a different object.

api/clusterlink.net/v1alpha1/operator_types.go
```diff
@@ -0,0 +1,122 @@
...
+	Status  StatusType `json:"status,omitempty"`
+	Message string     `json:"message,omitempty"`
+}
+
+// DataPlaneSpec defines the desired state of the dataplane components in ClusterLink.
+type DataPlaneSpec struct {
+	// +kubebuilder:validation:Enum=envoy;go
+	// +kubebuilder:default=envoy
+	// Type represents the dataplane type. Supports values "go" and "envoy".
+	Type string `json:"type,omitempty"`
```
orozery, line 83: Can you define `DataplaneType`, similar to `IngressType`?

api/clusterlink.net/v1alpha1/operator_types.go
```diff
@@ -0,0 +1,122 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   OperatorSpec   `json:"spec,omitempty"`
+	Status OperatorStatus `json:"status,omitempty"`
+}
+
+//+kubebuilder:object:root=true
+
+// OperatorList represents a list of Operator objects.
+type OperatorList struct {
```
orozery, line 114: I still don't see the reason for defining this struct if we don't plan to use it.

orozery (changes requested):

api/clusterlink.net/v1alpha1/operator_types.go
kfirtoledo: In all the examples I saw there is always a list object.
Also, I tried to remove it and got an error because the listKind always fills with a default value:
https://docs.openshift.com/container-platform/4.15/rest_api/extension_apis/customresourcedefinition-apiextensions-k8s-io-v1.html

orozery: Indeed I see it is defined by convention. Please ignore my previous comment :)

pkg/apis/clusterlink.net/v1alpha1/doc.go
elevran, line 1: Is this autogenerated as well?
It's an empty file. Typically, `doc.go` is used for package documentation...

pkg/apis/clusterlink.net/v1alpha1/operator_types.go
```diff
@@ -0,0 +1,131 @@
...
+// IngressType represents the ingress type  of the deployed ClusterLink.
+type IngressType string
+
+const (
+	// IngressTypeNone indicates that the deployment operator did not create any external ingress.
+	IngressTypeNone IngressType = "none"
+	// IngressTypeNodePort indicates that the deployment operator created an external ingress of type NodePort.
+	IngressTypeNodePort IngressType = "NodePort"
+	// IngressTypeLoadBalancer indicates that the deployment operator created an external ingress of type LoadBalancer.
+	IngressTypeLoadBalancer IngressType = "LoadBalancer"
```
elevran, line 43: Other types (future?) may be "gateway" (for Gateway API) or "ingress" (for current Ingress API)?

pkg/apis/clusterlink.net/v1alpha1/operator_types.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+// DataplaneType represents the dataplane type  of the deployed ClusterLink.
+type DataplaneType string
+
+const (
+	// GoType indicates that the dataplane type is Go-dataplane.
+	GoType DataplaneType = "go"
```
elevran, line 51: to be consistent with other types consider ` DataplaneTypeGo` and ` DataplaneTypeEnvoy`?

pkg/apis/clusterlink.net/v1alpha1/operator_types.go
```diff
@@ -0,0 +1,131 @@
...
+	// EnvoyType indicates that the dataplane type is Envoy.
+	EnvoyType DataplaneType = "envoy"
+)
+
+// OperatorSpec defines the desired state of a ClusterLink Operator.
```
elevran, line 56: I think it defines the desired state of the ClusterLink **deployment** (or **instance**), not the operator. The operator is a tool, not an end goal. The end goal is a deployed ClusterLink instance.
Please reconsider the use of `Operator` as a concept exposed to the developers, so `DeploymentSpec` and `DeploymentStatus` not `OperatorSpec`, and `OperatorStatus`, etc.

pkg/apis/clusterlink.net/v1alpha1/operator_types.go
```diff
@@ -0,0 +1,131 @@
...
+type OperatorSpec struct {
+	DataPlane DataPlaneSpec `json:"dataplane,omitempty"`
+	Ingress   IngressSpec   `json:"ingress,omitempty"`
+
+	// +kubebuilder:validation:Enum=trace;debug;info;warning;error;fatal
+	// +kubebuilder:default=info
+	// LogLevel define the ClusterLink components log level.
```
elevran, line 63:
```suggestion
	// LogLevel defines the ClusterLink components log level.
```

pkg/apis/clusterlink.net/v1alpha1/operator_types.go
```diff
@@ -0,0 +1,131 @@
...
+	LogLevel string `json:"logLevel,omitempty"`
+
+	// +kubebuilder:default="ghcr.io/clusterlink-net"
+	// ContainerRegistry is the container registry to pull the ClusterLink project images.
```
elevran, line 67:
```suggestion
	// ContainerRegistry is the container registry to pull the ClusterLink images.
```

pkg/apis/clusterlink.net/v1alpha1/operator_types.go
```diff
@@ -0,0 +1,131 @@
...
+	ContainerRegistry string `json:"containerRegistry,omitempty"`
+	// +kubebuilder:default="latest"
+	// ImageTag represents the version of the ClusterLink project images.
+	ImageTag string `json:"imageTag,omitempty"`
```
elevran, line 71: will we update a running deployment of the tag is changed on an existing spec? I think users expect that (redeploy) behavior.
If not, document that this is a one time deployment and users should delete/recreate on every change. 
IIRC, there is a kubebuilder [plugin](https://book.kubebuilder.io/plugins/deploy-image-plugin-v1-alpha) to help with it. Please check if useful/appropriate for our use case.

pkg/apis/clusterlink.net/v1alpha1/operator_types.go
```diff
@@ -0,0 +1,131 @@
...
+// ComponentStatus defines the status of component in ClusterLink.
+type ComponentStatus struct {
+	// +kubebuilder:validation:Enum=none;ready;error;pending
+	// +kubebuilder:default=none
+	Status  StatusType `json:"status,omitempty"`
+	Message string     `json:"message,omitempty"`
+}
+
+// DataPlaneSpec defines the desired state of the dataplane components in ClusterLink.
+type DataPlaneSpec struct {
```
elevran, line 89: Especially since you have status for both, Is there a corresponding `ControlPlaneSpec`?

pkg/apis/clusterlink.net/v1alpha1/operator_types.go
```diff
@@ -0,0 +1,131 @@
...
+type OperatorStatus struct {
+	Controlplane ComponentStatus `json:"controlplane,omitempty"`
+	Dataplane    ComponentStatus `json:"dataplane,omitempty"`
+}
+
+// ComponentStatus defines the status of component in ClusterLink.
+type ComponentStatus struct {
+	// +kubebuilder:validation:Enum=none;ready;error;pending
+	// +kubebuilder:default=none
+	Status  StatusType `json:"status,omitempty"`
```
elevran, line 84: Consider if `Conditions` are applicable?

pkg/apis/clusterlink.net/v1alpha1/operator_types.go
```diff
@@ -0,0 +1,131 @@
...
+	// Replicas represents the number of dataplane replicas.
+	Replicas int `json:"replicas,omitempty"`
+}
+
+// IngressSpec defines the type of the ingress component in ClusterLink.
+type IngressSpec struct {
+	// +kubebuilder:validation:Enum=none;LoadBalancer;NodePort
+	// +kubebuilder:default=none
+	// Type represents the type of service used to expose the ClusterLink deployment. Supported values: "LoadBalancer","NodePort", "none".
+	Type IngressType `json:"type,omitempty"`
```
elevran, line 106: what will be the name generated for the LoadBalancer of NodePort? User configured, or not? what are the defaults

elevran (changes requested):

pkg/apis/clusterlink.net/v1alpha1/doc.go
kfirtoledo: No. @orozery request to add it, I can remove it

pkg/apis/clusterlink.net/v1alpha1/operator_types.go
kfirtoledo: Yes we can add them,but I think in the moment we need them currently in all our demos we use only nodeport or loasvalancer

pkg/apis/clusterlink.net/v1alpha1/operator_types.go
kfirtoledo: I updated the comment in the types

pkg/apis/clusterlink.net/v1alpha1/operator_types.go
kfirtoledo: For now I don't need it, but probably with the new controlplane I will add it.

pkg/apis/clusterlink.net/v1alpha1/instance_types.go
```diff
@@ -0,0 +1,126 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// StatusConditionType represents the status conditions type for ClusterLink components.
```
elevran, line 20: nit: in the future consider grouping related (e.g., Status, Spec) structures together. Also, consider if defining top-down or bottom up (i.e., is the contained struct defined before or after its containing struct).

pkg/apis/clusterlink.net/v1alpha1/instance_types.go
```diff
@@ -0,0 +1,126 @@
...
+	// LogLevel define the ClusterLink components log level.
+	LogLevel string `json:"logLevel,omitempty"`
+
+	// +kubebuilder:default="ghcr.io/clusterlink-net"
+	// ContainerRegistry is the container registry to pull the ClusterLink project images.
+	ContainerRegistry string `json:"containerRegistry,omitempty"`
+	// +kubebuilder:default="latest"
+	// ImageTag represents the version of the ClusterLink project images.
+	ImageTag string `json:"imageTag,omitempty"`
+}
```
elevran, line 67: future: add a destination namespace (for now it's fixed to `clusterlink-system`) but may want to have user control over that.
You can assume the namespace must exist before (e.g., created by the user when/before the operator is deployed) and fail if it does not.

pkg/apis/clusterlink.net/v1alpha1/instance_types.go
```diff
@@ -0,0 +1,126 @@
...
+//+kubebuilder:object:root=true
+
+// InstanceList represents a list of Instance objects.
+type InstanceList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []Instance `json:"items"`
+}
+
+func init() {
```
elevran, line 124: question: is this where kubebuilder places the `init()` or is it copied from a different file?
I was thinking it belonged in the `groupversion_info.go` file...

elevran (approved):

pkg/apis/clusterlink.net/v1alpha1/instance_types.go
kfirtoledo: done by the kubebuilder, I just added the instance structs
https://kubebuilder.io/cronjob-tutorial/new-api#:~:text=Finally%2C%20we%20add%20the%20Go%20types%20to%20the%20API%20group.%20This%20allows%20us%20to%20add%20the%20types%20in%20this%20API%20group%20to%20any%20Scheme.

pkg/apis/clusterlink.net/v1alpha1/instance_types.go
kfirtoledo: You are absolutely right, I usually do top-down, the Kubebuilder convention is button-up, and that's what case the mix in the definition.

orozery (approved):

pkg/apis/clusterlink.net/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,33 @@
+// Copyright 2023 The ClusterLink Authors.
```
elevran, line 1: Nit: It's 2024, ...
Need to update the scripts (or the codegen boilerplate under kubebuilder's `hack` directory) to handle that correctly.
Should be done as separate issue/PR #285

kfirtoledo:
@elevran, @orozery PTAL

PR #261, opened 2024-01-11
Makefile: Add codegen target

kfirtoledo:
Add controller-gen target, codegen target and its outputs:
1. `clusterlink.net_operators.yaml` - The CRD yaml used by the operator.
2. `rbac/role.yaml` - The RBACs for the operator.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,26 @@ GO ?= CGO_ENABLED=0 go
 # Allow setting of go build flags from the command line.
 GOFLAGS := 
 
+# Location to install dependencies to
+GOBIN ?= $(GOPATH)/bin
+
+
+# Controller Gen for crds
+CONTROLLER_GEN ?= $(GOBIN)/controller-gen
+CONTROLLER_TOOLS_VERSION ?= v0.13.0
+
+controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary. If wrong version is installed, it will be overwritten.
+$(CONTROLLER_GEN):
+	test -s $(GOBIN)/controller-gen && $(GOBIN)/controller-gen --version | grep -q $(CONTROLLER_TOOLS_VERSION) || \
+	go install sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_TOOLS_VERSION)
+
+.PHONY: codegen
+codegen: controller-gen  ## Generate ClusterRole, CRDs and DeepCopyObject.
+	$(CONTROLLER_GEN) crd paths="./pkg/apis/..." output:crd:artifacts:config=config/operator/crds/
+	$(CONTROLLER_GEN) rbac:roleName=cl-operator-manager-role paths="./pkg/operator/..." output:rbac:dir=config/operator/rbac
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="././pkg/apis/..."
+	@goimports -l -w ./pkg/apis/clusterlink.net/v1alpha1/zz_generated.deepcopy.go
+
 build:
 	@echo "Start go build phase"
 	$(GO) build -o ./bin/gwctl ./cmd/gwctl
--- a/config/operator/rbac/role.yaml
+++ b/config/operator/rbac/role.yaml
@@ -1,0 +1,96 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: cl-operator-manager-role
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - nodes
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - persistentvolumeclaims
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
+  - services
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - apps
+  resources:
+  - deployments
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - clusterlink.net
+  resources:
+  - instances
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -71,6 +71,26 @@ GO ?= CGO_ENABLED=0 go
...
+CONTROLLER_TOOLS_VERSION ?= v0.13.0
+
+controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary. If wrong version is installed, it will be overwritten.
+$(CONTROLLER_GEN): $(LOCALBIN)
+	test -s $(LOCALBIN)/controller-gen && $(LOCALBIN)/controller-gen --version | grep -q $(CONTROLLER_TOOLS_VERSION) || \
+	GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_TOOLS_VERSION)
+
+.PHONY: codegen
+codegen: controller-gen  ## Generate ClusterRole, CRDs and DeepCopyObject.
+	$(CONTROLLER_GEN) crd paths="./api/..." output:crd:artifacts:config=config/operator/crd/bases
```
orozery, line 90: We still need to get closure on this (i.e. `pkg/apis`)

orozery: I'm wondering if we should create a separate output folder for the operator.
Maybe just use `config/crds`? 

Another option is `config/{core,operator}/{crds, rbac}`

Makefile
```diff
@@ -71,6 +71,26 @@ GO ?= CGO_ENABLED=0 go
...
+	$(CONTROLLER_GEN) rbac:roleName=cl-operator-manager-role paths="./pkg/operator/..." output:rbac:dir=config/operator/rbac
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
```
orozery, line 92: Shouldn't path be set just to the `api` (`pkg/apis`) directory?

Makefile
```diff
@@ -71,6 +71,26 @@ GO ?= CGO_ENABLED=0 go
 # Allow setting of go build flags from the command line.
 GOFLAGS := 
 
+# Location to install dependencies to
+LOCALBIN ?= $(shell pwd)/bin
+$(LOCALBIN):
```
elevran, line 76: is controller-gen a go program? Can it also be found in the Go path bin directory?

Makefile
```diff
@@ -71,6 +71,26 @@ GO ?= CGO_ENABLED=0 go
...
+	mkdir -p $(LOCALBIN)
+
+# Controller Gen for crds
+CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
+CONTROLLER_TOOLS_VERSION ?= v0.13.0
+
+controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary. If wrong version is installed, it will be overwritten.
+$(CONTROLLER_GEN): $(LOCALBIN)
+	test -s $(LOCALBIN)/controller-gen && $(LOCALBIN)/controller-gen --version | grep -q $(CONTROLLER_TOOLS_VERSION) || \
+	GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_TOOLS_VERSION)
```
elevran, line 86: GOBIN is likely already defined for the user. why not install there?

Makefile
```diff
@@ -71,6 +71,26 @@ GO ?= CGO_ENABLED=0 go
...
+
+.PHONY: codegen
+codegen: controller-gen  ## Generate ClusterRole, CRDs and DeepCopyObject.
+	$(CONTROLLER_GEN) crd paths="./pkg/apis/..." output:crd:artifacts:config=config/operator/crds/
+	$(CONTROLLER_GEN) rbac:roleName=cl-operator-manager-role paths="./pkg/operator/..." output:rbac:dir=config/operator/rbac
```
elevran, line 91: as in other PR's, I suggest not using `operator` as a concept, directory ,...

config/operator/crds/clusterlink.net_operators.yaml
```diff
@@ -0,0 +1,133 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.13.0
+  name: operators.clusterlink.net
```
elevran, line 7: ditto: use of `operator`

config/operator/rbac/role.yaml
```diff
@@ -0,0 +1,84 @@
...
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - ""
+  resources:
```
elevran, line 24: why do we need a Persistent volume claim?

config/operator/rbac/role.yaml
```diff
@@ -0,0 +1,84 @@
...
+  - clusterlink.net
+  resources:
+  - operators/status
+  verbs:
+  - get
+  - patch
+  - update
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
```
elevran, line 74: do we really need these?

config/operator/rbac/role.yaml
```diff
@@ -0,0 +1,84 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: cl-operator-manager-role
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - endpoints
+  - namespaces
```
elevran, line 11: seems to be overly permissive?
For example, why do we need to create namespaces, endpoints, etc. at runtime?

elevran: I understand this is autogenerated so should be fixed at the source, where the annotations are defined.

elevran (changes requested):

config/operator/rbac/role.yaml
kfirtoledo: This is how the controlplane is implemented for mapping the K8s secrets.

Makefile
kfirtoledo: I took this code from the Kubebuilder Makefile. Anyway, in my environment, it is not defined, and in the case the user doesn't define it.

Makefile
kfirtoledo: This is the same code as the Kubebuilder Makefile, but I think because it is related to our installation it should be in our local bin.

elevran (approved):

PR #262, opened 2024-01-11
config/operator: Add operator YAMLs

kfirtoledo:
Add operator deployment(manager) and rbacs.
Note: All the config files were generated by the Kubebuildrer CLI.

References #286

### DIFF
--- a/config/operator/manager/manager.yaml
+++ b/config/operator/manager/manager.yaml
@@ -1,0 +1,82 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
+    app.kubernetes.io/name: namespace
+    app.kubernetes.io/instance: clusterlink-operator
+    app.kubernetes.io/component: manager
+    app.kubernetes.io/created-by: cl-operator
+    app.kubernetes.io/part-of: cl-operator
+  name: clusterlink-operator
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app.kubernetes.io/component: manager
+    app.kubernetes.io/created-by: cl-operator
+    app.kubernetes.io/instance: controller-manager
+    app.kubernetes.io/name: deployment
+    app.kubernetes.io/part-of: cl-operator
+    control-plane: controller-manager
+  name: cl-operator-controller-manager
+  namespace: clusterlink-operator
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      control-plane: controller-manager
+  template:
+    metadata:
+      annotations:
+        kubectl.kubernetes.io/default-container: manager
+      labels:
+        control-plane: controller-manager
+    spec:
+      containers:
+      - args:
+        - --leader-elect
+        command:
+        - /manager
+        image: cl-operator:latest
+        imagePullPolicy: IfNotPresent
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8081
+          initialDelaySeconds: 15
+          periodSeconds: 20
+        name: manager
+        readinessProbe:
+          httpGet:
+            path: /readyz
+            port: 8081
+          initialDelaySeconds: 5
+          periodSeconds: 10
+        resources:
+          limits:
+            cpu: 500m
+            memory: 128Mi
+          requests:
+            cpu: 10m
+            memory: 64Mi
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+            - ALL
+      securityContext:
+        runAsNonRoot: true
+      serviceAccountName: cl-operator-controller-manager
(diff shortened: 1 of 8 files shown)


config/operator/manager/manager.yaml
```diff
@@ -0,0 +1,97 @@
...
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
+    app.kubernetes.io/name: namespace
+    app.kubernetes.io/instance: clusterlink-operator-ns
+    app.kubernetes.io/component: manager
+    app.kubernetes.io/created-by: cl-operator
+    app.kubernetes.io/part-of: cl-operator
+  name: clusterlink-operator-ns
```
elevran, line 11: is the choice of adding `-ns` Kubebuilder or input from the user?

config/operator/manager/manager.yaml
```diff
@@ -0,0 +1,97 @@
...
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app.kubernetes.io/component: manager
```
elevran, line 17: which of these (presence of key, the value assigned) are under our control, if any?

config/operator/manager/manager.yaml
```diff
@@ -0,0 +1,97 @@
...
+      labels:
+        control-plane: controller-manager
+    spec:
+      containers:
+      - args:
+        - --secure-listen-address=0.0.0.0:8443
+        - --upstream=http://127.0.0.1:8080/
+        - --logtostderr=true
+        - --v=0
+        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.15.0
```
elevran, line 43: How is this container used?
What does the image do in the context of our work?

config/operator/rbac/auth_proxy_client_clusterrole.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
elevran, line 1: What is the `auth_porxy` and how relevant to our use case? 
Could it be something that kubebuilder creates for other use cases? For example for an Authorization hook or something else?

I think we need to understand what these YAMLs are for and only generate/include the ones needed by us.
IMHO, there are more objects than we really need.

config/operator/rbac/operator_editor_role.yaml
```diff
@@ -0,0 +1,32 @@
+# permissions for end users to edit operators.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
elevran, line 3: the `clusterlink.net` objects are **namespaced**, thus I think we should have a Role and RoleBinding not cluster scoped roles and bindings.

config/operator/manager/manager.yaml
```diff
@@ -0,0 +1,97 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
+    app.kubernetes.io/name: namespace
+    app.kubernetes.io/instance: clusterlink-operator-ns
```
elevran, line 7: same `operator` comment as on the other PR's

elevran (changes requested):

config/operator/manager/manager.yaml
kfirtoledo: I added it, I can remove it...

config/operator/rbac/operator_editor_role.yaml
kfirtoledo: I will change it in the next phase, when we have the new controller

elevran (approved):

kfirtoledo:
@elevran PTAL

config/operator/manager/manager.yaml
```diff
@@ -0,0 +1,82 @@
...
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    app.kubernetes.io/name: namespace
+    app.kubernetes.io/instance: clusterlink-system
+    app.kubernetes.io/created-by: cl-operator
+    app.kubernetes.io/part-of: cl-operator
+  name: clusterlink-system
```
elevran, line 82: nit: GH complains of missing new line at EoF

config/operator/rbac/role_binding.yaml
```diff
@@ -0,0 +1,21 @@
...
+  namespace: clusterlink-operator
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cl-operator-manager-role
+subjects:
+- kind: ServiceAccount
+  name: cl-operator-controller-manager
+  namespace: clusterlink-operator
+
```
elevran, line 21: nit: EoL at EoF

elevran (approved):

Issue #61, opened 2023-10-03
Enable additional linters


elevran:
The current `golangci.yaml` configuration has some linters disabled as existing code does not pass them.
These should be enabled over time.

To enable a linter:
1. Set `issues.new: false` to run the linter on existing code (currently set to `true` to check only new code)
2. Uncomment relevant linter under `linters.enable` and/or the linter's configuration under `linters-settings.<name of linter>`
3. Run `make lint` and fix all reported errors
4. Revert `issues.new: true` and then commit the modified code and `golangci.yaml`

For additional information and reference you may want to refer to [this blog](https://freshman.tech/linting-golang/) and the golanci linters documentation on [linters](https://golangci-lint.run/usage/linters/) and their [configuration](https://golangci-lint.run/usage/configuration/)

elevran:
remaining linters:
| linter       | config | errors |
| ------      | ------- | ------- |
| gofumpt |            | ~55 |
| lll            |            | ~45 |
| nilnil       |            | ~10 |
| noctx      |            | ~5 |
| thelper    |           | ~5 |
| testpackage |      | ~5 |
| varnamelen |      | ~130 |
| gocritic | experimental | ~30 |
| gocritic | opinionated | ~45 |
| gocritic | style | ~45 |
| gocritic | hugeParam | ~5 |
| gocritic | rangeExprCopy | <5 |
| gocritic | rangeValCopy | <5 |
| errcheck | type-assert | ~10 |
| errcheck | blank | ~10 |

elevran:
Split remaining linters into individual issues (#270, #271, #272).
This will be closed once #269 is merged

PR #266, opened 2024-01-15
Enable `nestif` linter

elevran:
Replaces previous PR #250.
- simplify owner-ref logic, align happy path to left
- simplify reconcile logic, align happy path to left
- enable nestif linter in configuration

Reference #61


kfirtoledo (approved):
LGTM

PR #269, opened 2024-01-18
changes for gocritic linter

### LINKED ISSUE #61: Enable additional linters

elevran:
Fix #61

### DIFF
--- a/cmd/cl-controlplane/app/server.go
+++ b/cmd/cl-controlplane/app/server.go
@@ -129,13 +129,13 @@ func (o *Options) Run() error {
 
 	storeManager := kv.NewManager(kvStore)
 
-	// initiailze platform
-	var platform platform.Platform
+	// initialize platform
+	var pp platform.Platform
 	switch o.Platform {
 	case platformUnknown:
-		platform = unknown.NewPlatform()
+		pp = unknown.NewPlatform()
 	case platformK8S:
-		platform, err = k8s.NewPlatform()
+		pp, err = k8s.NewPlatform()
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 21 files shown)


kfirtoledo (approved):

cmd/cl-controlplane/app/server.go
```diff
@@ -129,21 +129,21 @@ func (o *Options) Run() error {
...
 	storeManager := kv.NewManager(kvStore)
 
-	// initiailze platform
-	var platform platform.Platform
+	// initialize platform
+	var pp platform.Platform
 	switch o.Platform {
 	case platformUnknown:
-		platform = unknown.NewPlatform()
+		pp = unknown.NewPlatform()
```
elevran, line 136: use of `platform` as the variable name shadows the import of `platform` package.
In most cases this won't be an issue and would produce correct code. Potential ambiguity arises when both the  variable and package have the same function name (i.e., should it call exported package function or function with struct as the method receiver?)

PR #273, opened 2024-01-18
variable name length linting fixes

### LINKED ISSUE #272: Enable `varnamelen` linter
See #61 for details.

elevran:
Fix #272

### DIFF
--- a/pkg/util/log/log.go
+++ b/pkg/util/log/log.go
@@ -30,7 +30,8 @@ const (
 
 // SetLog sets logrus logger (format, file, level).
 func SetLog(logLevel, logFileName string) (*os.File, error) {
-	var f *os.File
+	var logfile *os.File
+
 	if logFileName != "" {
 		usr, err := user.Current()
 		if err != nil {
@@ -63,7 +64,7 @@ func SetLog(logLevel, logFileName string) (*os.File, error) {
 			DisableQuote:    true,
 		},
 	})
-	return f, nil
+	return logfile, nil
 }
 
 type formatter struct {
--- a/pkg/util/log/log.go
+++ b/pkg/util/log/log.go
@@ -40,7 +40,7 @@ func SetLog(logLevel, logFileName string) (*os.File, error) {
 		logFileFullPath := path.Join(usr.HomeDir, logFileName)
 		createLogFolder(logFileFullPath)
 
-		logfile, err := os.OpenFile(logFileFullPath, os.O_APPEND|os.O_CREATE|os.O_RDWR, 0o600)
+		logfile, err = os.OpenFile(logFileFullPath, os.O_APPEND|os.O_CREATE|os.O_RDWR, 0o600)
 		fmt.Printf("Creating log file: %v\n", logFileFullPath)
 		if err != nil {
 			return nil, fmt.Errorf("error opening log file: %w", err)
--- a/pkg/util/rest/server.go
+++ b/pkg/util/rest/server.go
@@ -249,28 +249,28 @@ func (s *Server) list(spec *ServerObjectSpec, w http.ResponseWriter, r *http.Req
 
 // AddObjectHandlers adds the server a handlers for managing a specific object type.
 func (s *Server) AddObjectHandlers(spec *ServerObjectSpec) {
-	r := s.Router()
+	router := s.Router()
 
-	r.Route(spec.BasePath, func(r chi.Router) {
-		r.Post("/", func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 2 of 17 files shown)


pkg/util/rest/server.go
```diff
@@ -249,28 +249,28 @@ func (s *Server) list(spec *ServerObjectSpec, w http.ResponseWriter, r *http.Req
 
 // AddObjectHandlers adds the server a handlers for managing a specific object type.
 func (s *Server) AddObjectHandlers(spec *ServerObjectSpec) {
-	r := s.Router()
+	router := s.Router()
 
-	r.Route(spec.BasePath, func(r chi.Router) {
-		r.Post("/", func(w http.ResponseWriter, r *http.Request) {
+	router.Route(spec.BasePath, func(cr chi.Router) {
```
elevran, line 254: In the original form, this included 3 (!) variables named `r`:
- the chi.Router associated with the server (renamed: `router`);
- the chi.Router variable passed as the (unnamed) function parameter (renamed: `cr`); and
- the http.Request variable passed to handlers (unchanged)

pkg/util/log/log.go
```diff
@@ -30,7 +30,8 @@ const (
 
 // SetLog sets logrus logger (format, file, level).
 func SetLog(logLevel, logFileName string) (*os.File, error) {
-	var f *os.File
+	var logfile *os.File
```
zivnevo, line 33: Something seems wrong here. Was `f` shadowed in the inner scope?

elevran: no. It was a bad fix on my part in #269, where I had missed the setting of `f` inside the following block. Since I added `logfile` *and* `:=` in the inner block, `f` was never set to the intended return value in `os.OpenFile()`.
Pushing a fix...

zivnevo (approved):
LGTM

orozery (commented):
I would use `peer` instead of `pr`.

elevran:
> I would use `peer` instead of `pr`.

In tests/e2e/k8s/util/fabric.go? 
the concern was that it would shadow the object type (we would have `peer *peer` as parameter which compiles but is a matter of style preference) and "stuttering" (e.g., `peer.peerCert`).
@orozery Maybe `target` would be a good compromise (see below for snippets):
```go
func (f *Fabric) deployClusterLink(target *peer, cfg *PeerConfig) (*ClusterLink, error) { 
        ... 
        if err := target.cluster.CreateFromYAML(k8sYAML, f.namespace); err != nil { 
                 ... 
        }
        ...
        var service v1.Service
	err = target.cluster.resources.Get(context.Background(), "cl-dataplane", f.namespace, &service)
        ...
	c := client.New(target.cluster.IP(), port, &tls.Config{
		MinVersion:   tls.VersionTLS12,
		Certificates: []tls.Certificate{certificate},
		RootCAs:      caCertPool,
		ServerName:   target.cluster.Name(),
	})
        ...
}
```

orozery:
> > I would use `peer` instead of `pr`.
> 
> In tests/e2e/k8s/util/fabric.go? the concern was that it would shadow the object type (we would have `peer *peer` as parameter which compiles but is a matter of style preference) and "stuttering" (e.g., `peer.peerCert`).
> 
> two options:
> 
> 1. maybe `target` would be a good compromise (see below for snippets):
> 
> ```go
> func (f *Fabric) deployClusterLink(target *peer, cfg *PeerConfig) (*ClusterLink, error) { 
>         ... 
>         if err := target.cluster.CreateFromYAML(k8sYAML, f.namespace); err != nil { 
>                  ... 
>         }
>         ...
>         var service v1.Service
> 	err = target.cluster.resources.Get(context.Background(), "cl-dataplane", f.namespace, &service)
>         ...
> 	c := client.New(target.cluster.IP(), port, &tls.Config{
> 		MinVersion:   tls.VersionTLS12,
> 		Certificates: []tls.Certificate{certificate},
> 		RootCAs:      caCertPool,
> 		ServerName:   target.cluster.Name(),
> 	})
>         ...
> }
> ```
> 
> 2. ~keep `p *peer` _and_ introduce an intermediate variable for `cluster := p.cluster` which would also satisfy the name length requirements.~ Does not work as `p` is still to short compared with where used (defined-used distance is ~30)
> 
> @orozery regarding option (1) - WDYT?

(1) is ok, but it's ok to name a variable the same name as its type. It does not shadow anything. We actually do this in other places.

elevran:
> (1) is ok, but it's ok to name a variable the same name as its type. It does not shadow anything. We actually do this in other places.

Then option (1) it shall be ;-).

It is indeed syntactically correct, but I would prefer not to use it if possible (personal preference...)

PR #274, opened 2024-01-18
cmd/cl-operator: Introduce the deployment operator (cl-operator).

kfirtoledo:
The deployment operator creates an operator to deploy ClusterLink. 
Its main component is the instance_controller, which takes the ClusterLink instance CRD and creates the ClusterLink components
References #286 

**Note: This code depends on PR https://github.com/clusterlink-net/clusterlink/pull/259**

### DIFF
--- a/cmd/cl-operator/Dockerfile
+++ b/cmd/cl-operator/Dockerfile
@@ -1,0 +1,8 @@
+FROM alpine:3.14
+
+FROM gcr.io/distroless/static:nonroot
+WORKDIR /
+COPY  ./bin/manager .
+USER 65532:65532
+
+ENTRYPOINT ["/manager"]
--- a/cmd/cl-operator/main.go
+++ b/cmd/cl-operator/main.go
@@ -1,0 +1,126 @@
+// Copyright 2023 The ClusterLink Authors.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/*
+Copyright 2024.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"flag"
+	"os"
+
+	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
+	// to ensure that exec-entrypoint and run can make use of them.
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
+
+	clusterlink "github.com/clusterlink-net/clusterlink/pkg/apis/clusterlink.net/v1alpha1"
+	"github.com/clusterlink-net/clusterlink/pkg/operator/controller"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	scheme   = runtime.NewScheme()
+	setupLog = ctrl.Log.WithName("setup")
+)
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(clusterlink.AddToScheme(scheme))
+}
+
+func main() {
+	var metricsAddr string
+	var enableLeaderElection bool
+	var probeAddr string
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	flag.BoolVar(&enableLeaderElection, "leader-elect", false, "Enable leader election for controller manager. "+
+		"Enabling this will ensure there is only one active controller manager.")
+	opts := zap.Options{
+		Development: true,
+	}
+	opts.BindFlags(flag.CommandLine)
+	flag.Parse()
+
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme:                 scheme,
+		Metrics:                metricsserver.Options{BindAddress: metricsAddr},
+		HealthProbeBindAddress: probeAddr,
+		LeaderElection:         enableLeaderElection,
+		LeaderElectionID:       "2754cddd.clusterlink.net",
+		// LeaderElectionReleaseOnCancel defines if the leader should step down voluntarily
+		// when the Manager ends. This requires the binary to immediately end when the
+		// Manager is stopped, otherwise, this setting is unsafe. Setting this significantly
+		// speeds up voluntary leader transitions as the new leader don't have to wait
+		// LeaseDuration time first.
+		//
+		// In the default scaffold provided, the program ends immediately after
+		// the manager stops, so would be fine to enable this option. However,
+		// if you are doing or is intended to do any operation such as perform cleanups
+		// after the manager stops then its usage might be unsafe.
+		// LeaderElectionReleaseOnCancel: true,
+	})
+	if err != nil {
+		setupLog.Error(err, "unable to start manager")
+		os.Exit(1)
+	}
+
+	if err = (&controller.InstanceReconciler{
+		Client:        mgr.GetClient(),
+		Scheme:        mgr.GetScheme(),
+		Logger:        logrus.WithField("component", "reconciler"),
+		InstancesMeta: make(map[string]string),
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "Operator")
+		os.Exit(1)
+	}
+
+	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+		setupLog.Error(err, "unable to set up health check")
+		os.Exit(1)
+	}
+	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+		setupLog.Error(err, "unable to set up ready check")
+		os.Exit(1)
+	}
+
+	setupLog.Info("starting manager")
+	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
+		setupLog.Error(err, "problem running manager")
+		os.Exit(1)
+	}
+}
--- a/pkg/operator/controller/instance_controller.go
+++ b/pkg/operator/controller/instance_controller.go
@@ -1,0 +1,838 @@
+// Copyright 2023 The ClusterLink Authors.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package controller
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"reflect"
+	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+
+	cpapp "github.com/clusterlink-net/clusterlink/cmd/cl-controlplane/app"
+	dpapp "github.com/clusterlink-net/clusterlink/cmd/cl-dataplane/app"
+	clusterlink "github.com/clusterlink-net/clusterlink/pkg/apis/clusterlink.net/v1alpha1"
+	cpapi "github.com/clusterlink-net/clusterlink/pkg/controlplane/api"
+	dpapi "github.com/clusterlink-net/clusterlink/pkg/dataplane/api"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	ControlPlaneName  = "cl-controlplane"
+	DataPlaneName     = "cl-dataplane"
+	GoDataPlaneName   = "cl-go-dataplane"
+	IngressName       = "cl-external-svc"
+	OperatorNameSpace = "clusterlink-operator" // TODO -should be removed
+	InstanceNamespace = "clusterlink-system"
+	FinalizerName     = "instance.clusterlink.net/finalizer"
+
+	StatusModeNotExist    = "NotExist"
+	StatusModeProgressing = "ProgressingMode"
+	StatusModeReady       = "Ready"
+)
+
+// InstanceReconciler reconciles a ClusterLink instance object.
+type InstanceReconciler struct {
+	client.Client
+	Scheme        *runtime.Scheme
+	Logger        *logrus.Entry
+	InstancesMeta map[string]string
+}
+
+// +kubebuilder:rbac:groups=clusterlink.net,resources=instances,verbs=list;get;watch;update;patch
+// +kubebuilder:rbac:groups=clusterlink.net,resources=instances/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=clusterlink.net,resources=instances/finalizers,verbs=update
+// +kubebuilder:rbac:groups="",resources=persistentvolumeclaims,verbs=list;get;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=services;serviceaccounts,verbs=list;get;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=nodes,verbs=list;get;watch
+// +kubebuilder:rbac:groups="",resources=pods,verbs=list;get;watch
+// +kubebuilder:rbac:groups="apps",resources=deployments,verbs=list;get;watch;create;update;patch;delete
+//nolint:lll // Ignore long line warning for Kubebuilder command.
+// +kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=clusterroles;clusterrolebindings,verbs=list;get;watch;create;update;patch;delete
+
+// TODO- should review the operator RABCs.
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *InstanceReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&clusterlink.Instance{}).
+		Watches(
+			&appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: ControlPlaneName,
+				},
+			},
+			&handler.EnqueueRequestForObject{},
+		).
+		Watches(
+			&appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: DataPlaneName,
+				},
+			},
+			&handler.EnqueueRequestForObject{},
+		).
+		Complete(r)
+}
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// The reconcile get instance YAML and creates the ClusterLink components.
+func (r *InstanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// Fetch the Clusterlink instance
+	instance := &clusterlink.Instance{}
+	err := r.Get(ctx, req.NamespacedName, instance)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		// Requeue the request on error
+		return ctrl.Result{}, err
+	}
+
+	// Check one clusterlink per namespace
+	if name, exist := r.InstancesMeta[instance.Namespace]; exist {
+		if instance.Name != name {
+			err := fmt.Errorf("instance %s in Namespace %s is  already exist- Only one instance are permit in  a Namespace",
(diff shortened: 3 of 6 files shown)


pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+	r.Logger.Infof("Enter instance Reconcile - (Namespace: %s, Name: %s)", instance.Namespace, instance.Name)
+	r.Logger.Info("InstanceSpec- ",
+		" DataPlane.Type: ", instance.Spec.DataPlane.Type,
+		", DataPlane.Replicas: ", instance.Spec.DataPlane.Replicas,
+		", Ingress.Type: ", instance.Spec.Ingress.Type,
+		", LogLevel: ", instance.Spec.LogLevel,
+		", ContainerRegistry: ", instance.Spec.ContainerRegistry,
+		", ImageTag: ", instance.Spec.ImageTag,
+	)
+	r.instance = instance
```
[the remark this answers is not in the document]
elevran: It might be sufficient to just store a NamespacedName here, since you should have access to the Instance object and retrieve its fields on demand (it'll be cached in the controller-runtime). Having the instance namespace and name is also sufficient to mark additional instances as in conflict and provide the original object name.
I'm not a fan of copying and keeping a full copy on the reconciler.

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+	dpapi "github.com/clusterlink-net/clusterlink/pkg/dataplane/api"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	ControlPlaneName      = "cl-controlplane"
+	DataPlaneName         = "cl-dataplane"
+	GoDataPlaneName       = "cl-go-dataplane"
+	IngressName           = "cl-external-svc"
+	OperatorNameSpace     = "clusterlink-operator"
```
elevran, line 49: You should be able to read this from an environment variable and not hard code it.
```go
// Namespace returns the Pod namespace, either from environment or k8s files
func Namespace() (string, error) {
	const nsFile = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
	if ns := os.Getenv("POD_NAMESPACE"); ns != "" {
		return ns, nil
	}
	if data, err := ioutil.ReadFile(nsFile); err == nil {
		if ns := strings.TrimSpace(string(data)); len(ns) > 0 {
			return ns, nil
		}
		return "", err
	}
	return "", nil
}
```

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+		r.instance.Spec.ContainerRegistry += "/"
+	}
+
+	// Set the deployment namespace
+	r.clNamespace = InstanceNamespace
+	if r.instance.Namespace != OperatorNameSpace { // Allow to create in any namespace. TODO -should remove this part and use only InstanceNamespace
+		r.clNamespace = r.instance.Namespace
+	}
+
+	// Examine DeletionTimestamp to determine if object is under deletion
```
elevran, line 136: May be better to handle deletion as the first case to handle (e.g., around L114)

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+	if !r.instance.DeletionTimestamp.IsZero() {
+		if err := r.deleteAllResources(ctx); err != nil {
+			return ctrl.Result{}, err
+		}
+
+		if err := r.deleteFinalizer(ctx); err != nil {
+			return ctrl.Result{}, err
+		}
```
elevran, line 144: Consider clearing `r.instance` as well (maybe explicitly separate from `r.deleteAllResources`?).
After you delete the instance, you should be back to the original state.

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+	}
+	// Set Finalizer if needed
+	if err := r.setFinalizer(ctx); err != nil {
```
elevran, line 147: Does it override the finalizer if already exists?

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+		return ctrl.Result{}, err
+	}
+
+	// Set ClusterLink components if needed
+	if err := r.setClusterLink(ctx); err != nil {
```
elevran, line 152: IMO `create`,  `generate` (or `apply` if you want to be k8s specific...) are better verbs than `set`.
nit 1: I think the comment is redundant - it does exactly what its name says.
nit 2: since this function (`setClusterLink`) is meant to do the opposite of `deleteAllResources`,  I would request that their names reflect that (e.g., create and delete ClusterLink as an example)

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+	// Set deployments
+	if err := r.setControlplane(ctx); err != nil {
+		return err
+	}
+	if err := r.setDataplane(ctx); err != nil {
+		return err
+	}
+
+	// Set rules
+	if err := r.setRules(ctx, ControlPlaneName); err != nil {
```
elevran, line 198: what do `rules` refer to? RBAC is Role Based Access Control, so maybe use setRBAC or setAccessControl?

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+		r.instance.Status.Dataplane.Conditions[string(api.DeploymentReady)].Reason == StatusModeProgressing {
+		return ctrl.Result{Requeue: true, RequeueAfter: time.Second * 5}, err
+	}
+	return ctrl.Result{}, nil
+}
+
+// setClusterLink sets up all the components for the ClusterLink project.
+func (r *InstanceReconciler) setClusterLink(ctx context.Context) error {
+
+	// Set Namespace
```
elevran, line 171: ditto: set or create?
`set` seems to suggest variable assignment (set/get), not an operation

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{"app": ControlPlaneName},
+				},
+				Spec: corev1.PodSpec{
+					Volumes: []corev1.Volume{
+						{
+							Name: "ca",
+							VolumeSource: corev1.VolumeSource{
+								Secret: &corev1.SecretVolumeSource{
```
elevran, line 227: does it always look for secrets in the same namespace as the Pod or do we need to specify it? In k8s, empty namespace usually means `default` which is not what we want when using `clusterlink-operator` and/or `clusterlink-system`

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+									SecretName: "cl-fabric",
+								},
+							},
+						},
+						{
+							Name: "tls",
```
elevran, line 233: Is there a more descriptive name?
TLS of which component? Maybe certificates?

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+							Name:            ControlPlaneName,
+							Image:           r.instance.Spec.ContainerRegistry + ControlPlaneName + ":" + r.instance.Spec.ImageTag,
+							ImagePullPolicy: corev1.PullIfNotPresent,
+							Args:            []string{"--log-level", r.instance.Spec.LogLevel, "--platform", "k8s"},
+							Ports: []corev1.ContainerPort{
+								{
+									ContainerPort: cpapi.ListenPort,
+								},
+							},
+							VolumeMounts: []corev1.VolumeMount{
```
elevran, line 260: are the volumes needed on the deployment template and its containers spec?
Asking due to ignorance...

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+									ReadOnly:  true,
+								},
+								{
+									Name:      ControlPlaneName,
+									MountPath: filepath.Dir(cpapp.StoreFile),
+								},
+							},
+							Env: []corev1.EnvVar{
+								{
+									Name: "CL-NAMESPACE",
```
elevran, line 286: See earlier note on hardcoding the namespace name. It can be made available to the operator pods in the same manner as you provide it to the control plane.

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+}
+
+// setDataplane sets up the dataplane deployment.
+func (r *InstanceReconciler) setDataplane(ctx context.Context) error {
+	DataplaneImage := DataPlaneName
+	if r.instance.Spec.DataPlane.Type == api.DataplaneTypeGo {
+		DataplaneImage = GoDataPlaneName
+	}
+
+	dpDeployment := &appsv1.Deployment{
```
elevran, line 311: nit: Consider adding a function that sets all of the common fields in the two deployments and then overriding specific items in each. Might keep code shorter.
Or maybe just the Volume stuff can be extracted?

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+					Protocol: corev1.ProtocolTCP,
+					Port:     int32(port),
+				},
+			},
+			Type:     corev1.ServiceTypeClusterIP,
+			Selector: map[string]string{"app": name},
+		},
+	}
+
+	return r.createResource(ctx, service)
```
elevran, line 404: any reason not to use `createOrUpdateResource` here as done in the other places?

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+	return r.createResource(ctx, controlplanePVC)
+}
+
+// setRules sets up k8s ClusterRule and ClusterRoleBinding for the controlplane.
+func (r *InstanceReconciler) setRules(ctx context.Context, name string) error {
+	// Create the ClusterRole for the controlplane.
+	clusterRole := &rbacv1.ClusterRole{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: r.clNamespace,
```
elevran, line 434: q: are ClusterRole objects namespaced?

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+		},
+		Rules: []rbacv1.PolicyRule{
+			{
+				APIGroups: []string{""},
+				Resources: []string{"services"},
+				Verbs:     []string{"create", "delete", "update"},
```
elevran, line 440: watch? get?

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "ClusterRole",
+			Name:     name,
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      "default",
```
elevran, line 469: I don't think the control plane should be run with default service account in the default namespace. That's not where you deploy it, is it?
The default service account is usually without privileges so I would run the control plane with a SA that is explicit.

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+				Namespace: "default",
+			},
+		},
+	}
+	return r.createResource(ctx, clusterRoleBinding)
+}
+
+// setNameSpace sets up k8s ClusterRule and ClusterRoleBinding for the controlplane.
+func (r *InstanceReconciler) setNameSpace(ctx context.Context) error {
```
elevran, line 478: - Comment seems not aligned with the actual code.
- I think you can assume the destination namespace always exists and fail if it does not (mark the conditions on the Instance object to failed or invalid as well). I would consider adding the destination namespace to the Spec (default value is `clusterlink-system` when not set. It will make life easier and allow some customization to users without much work on our part.

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+		Spec: corev1.ServiceSpec{
+			Selector: map[string]string{
+				"app": "cl-dataplane",
+			},
+			Ports: []corev1.ServicePort{
+				{
+					Port:       dpapi.ListenPort,
+					TargetPort: intstr.FromInt(dpapi.ListenPort),
+					Protocol:   corev1.ProtocolTCP,
+					Name:       "http",
```
elevran, line 505: not https or tcp?

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+			},
+		},
+	}
+
+	switch r.instance.Spec.Ingress.Type {
+	case api.IngressTypeNone:
+		return nil
+	case api.IngressTypeNodePort:
+		service.Spec.Type = corev1.ServiceTypeNodePort
+		service.Spec.Ports[0].NodePort = ExternalNodePort
```
elevran, line 516: do we want to hard code the port or let k8s report it for us (and we can reflect in the status)?
If the port is taken - this will fail - in which case you need to reflect back to the Instance.Status.

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+					TargetPort: intstr.FromInt(dpapi.ListenPort),
+					Protocol:   corev1.ProtocolTCP,
+					Name:       "http",
+				},
+			},
+		},
+	}
+
+	switch r.instance.Spec.Ingress.Type {
+	case api.IngressTypeNone:
```
elevran, line 512: Move this up since there's no need to define/create the Service object at all in this case.

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+	if err := r.deleteResource(ctx, ControlPlaneName, &rbacv1.ClusterRoleBinding{}); err != nil {
+		return err
+	}
+
+	// Delete ingress
+	if err := r.deleteResource(ctx, IngressName, &corev1.Service{}); err != nil {
+		return err
+	}
+
+	if r.clNamespace == InstanceNamespace {
```
elevran, line 611: I think namespace creation/deletion should be out of scope for the operator.
Think about the case where we deploy into a user namespace (future, I know).

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+		}
+	}
+
+	return nil
+
+}
+
+// deleteResource delete a k8s resource.
+func (r *InstanceReconciler) deleteResource(ctx context.Context, name string, object client.Object) error {
+	err := r.Get(ctx, types.NamespacedName{Name: name, Namespace: r.clNamespace}, object)
```
elevran, line 623: If the intent is to delete always, why do you need a Get first?

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+}
+
+// Helper function to convert int32 to *int32
+func int32Ptr(i int32) *int32 {
+	return &i
+}
+
+// checkStatus check the status of ClusterLink components.
+func (r *InstanceReconciler) checkStatus(ctx context.Context) error {
+	// Delete ingress
```
elevran, line 638: comment seems wrong

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+}
+
+// checkDeploymnetStatus check the status of a deployment.
+func (r *InstanceReconciler) checkDeploymnetStatus(ctx context.Context, name types.NamespacedName) (metav1.Condition, error) {
+	d := &appsv1.Deployment{}
+	status := metav1.Condition{Type: string(api.DeploymentReady), Status: metav1.ConditionFalse, LastTransitionTime: metav1.Now()}
+	if err := r.Get(ctx, name, d); err != nil {
+		if errors.IsNotFound(err) {
+			status.Reason = "NotExist"
+			status.Message = "Deployment is not exist"
```
elevran, line 692:
```suggestion
			status.Message = "Deployment does not exist"
```

elevran (changes requested):

cmd/cl-operator/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM alpine:3.14
+
+FROM gcr.io/distroless/static:nonroot
+WORKDIR /
+COPY  ./bin/manager .
+USER 65532:65532
```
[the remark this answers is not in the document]
kfirtoledo: https://docs.docker.com/engine/reference/builder/#user:~:text=run%20the%20container.-,USER,-USER%20%3Cuser%3E%5B%3A%3Cgroup
The numbers 65532:65532 are often used for a non-root user

cmd/cl-operator/main.go
```diff
@@ -0,0 +1,130 @@
...
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
+
+	api "github.com/clusterlink-net/clusterlink/pkg/apis/clusterlink.net/v1alpha1"
+	"github.com/clusterlink-net/clusterlink/pkg/operator/controller"
+	"github.com/sirupsen/logrus"
+	//+kubebuilder:scaffold:imports
```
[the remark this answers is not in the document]
kfirtoledo: I will remove it.
It is part of the creation part of Kubebuilder to add imports to the main, we don't need it anymore.

cmd/cl-operator/main.go
```diff
@@ -0,0 +1,130 @@
...
+var (
+	scheme   = runtime.NewScheme()
+	setupLog = ctrl.Log.WithName("setup")
+)
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+
+	utilruntime.Must(api.AddToScheme(scheme))
+	//+kubebuilder:scaffold:scheme
```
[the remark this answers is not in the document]
kfirtoledo: Same as before I will remove it

cmd/cl-operator/main.go
```diff
@@ -0,0 +1,130 @@
...
+
+func main() {
+	var metricsAddr string
+	var enableLeaderElection bool
+	var probeAddr string
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
+		"Enable leader election for controller manager. "+
+			"Enabling this will ensure there is only one active controller manager.")
```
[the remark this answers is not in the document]
kfirtoledo: This is the default value the kubebuilder generated. I don't see a reason to change it, especially since we have only one controller  manager.

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+
+const (
+	ControlPlaneName      = "cl-controlplane"
+	DataPlaneName         = "cl-dataplane"
+	GoDataPlaneName       = "cl-go-dataplane"
+	IngressName           = "cl-external-svc"
+	OperatorNameSpace     = "clusterlink-operator"
+	InstanceNamespace     = "clusterlink-system"
+	FinalizerName         = "instance.clusterlink.net/finalizer"
+	ExternalNodePort      = 30443
```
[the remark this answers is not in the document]
kfirtoledo: For the nodeport external service, if we want we can allow the user to control it by adding the node port to the API.
Personally, I think it is not needed.

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+	StatusModeReady       = "Ready"
```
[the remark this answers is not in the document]
kfirtoledo: This is the reason field is in the status condition,  I think it is part of the implementation.

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+	StatusModeProgressing = "ProgressingMode"
```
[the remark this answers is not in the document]
kfirtoledo: This is only the reason why the DeploymentReady-condition is false - I prefer to be aligned like the deployment status.condition -https://github.com/kubernetes/kubernetes/blob/418ae605ec1b788d43bff7ac44af66d8b669b833/pkg/apis/apps/types.go#L556
For the case it is invalid, I will pass the deployment reason.

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+	Scheme      *runtime.Scheme
+	instance    *api.Instance
+	clNamespace string
+	Logger      *logrus.Entry
+	CaFabric    []byte
+}
+
+// +kubebuilder:rbac:groups=clusterlink.net,resources=instances,verbs=list;get;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=clusterlink.net,resources=instances/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=clusterlink.net,resources=instances/finalizers,verbs=update
```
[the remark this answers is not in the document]
kfirtoledo: It is in the metadata, but from what I read it is considered a resource.
I use the example in the kubebuilder and every time you delete or create the finalizers you use the update command.

pkg/operator/controller/instance_controller.go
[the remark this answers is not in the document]
kfirtoledo: This copy is only used in a single call of the reconciliation, I use this copy in part of the functions and also update the status,
I can change it to a variable and pass it to all the functions, but I prefer this way.

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+// +kubebuilder:rbac:groups="",resources=persistentvolumeclaims,verbs=list;get;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=services;namespaces,verbs=list;get;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=pods,verbs=list;get;watch
```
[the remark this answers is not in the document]
kfirtoledo: It is for the RBAC for the cl-controller (if the operator gives him permissions, he also need to have those permissions)

pkg/operator/controller/instance_controller.go
kfirtoledo: No, it checks if it already contains

pkg/operator/controller/instance_controller.go
kfirtoledo: I will remove it when we have the cl-controller (can create import service in different namespaces)
I will add TODO.

pkg/operator/controller/instance_controller.go
kfirtoledo: Yes, by default the secrets should be in the same namespace as the pods ( I can add it to the code but I think it is clear enough)

pkg/operator/controller/instance_controller.go
kfirtoledo: This part of the controlplane- @orozery can change it in the new controlplane.

pkg/operator/controller/instance_controller.go
kfirtoledo: Container,
From the code:
	// Pod volumes to mount into the container's filesystem.
	// Cannot be updated.
	VolumeMounts []VolumeMount `json:"volumeMounts,omitempty" patchStrategy:"merge"

pkg/operator/controller/instance_controller.go
kfirtoledo: Added to the issue of the operator.

pkg/operator/controller/instance_controller.go
```diff
@@ -0,0 +1,742 @@
...
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// The reconcile get instance YAML and creates the ClusterLink components.
+func (r *InstanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// Fetch the Clusterlink instance
+	instance := &api.Instance{}
+	err := r.Get(ctx, req.NamespacedName, instance)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
kfirtoledo: But I have a finalizer, unless someone deletes the finalizer from outside before deleting the CRD instance it will be ok.

pkg/operator/controller/instance_controller.go
kfirtoledo: Part of the controller (@orozery )

pkg/operator/controller/instance_controller.go
kfirtoledo: In the new controller we will need it In the last one it was reactive to the CLI. (I will update the in the issue https://github.com/clusterlink-net/clusterlink/issues/275)

pkg/operator/controller/instance_controller.go
kfirtoledo: To save API calls, and prevent an error in the delete command.

cmd/cl-operator/main.go
```diff
@@ -0,0 +1,127 @@
...
```
elevran, line 71: nit: please align indentation

elevran (approved):

Issue #17, opened 2023-09-21
Policy engine to consider more workload attributes when deciding on a connection


zivnevo:

elevran:
References #86

zivnevo:
Some implementation issues I'm currently struggling with:
1. There is currently no way to get the attributes of the requesting workload on the remote service side. We only know the name of the requesting peer. This means ingress connection requests cannot be evaluated the same way as the corresponding egress connection requests.
2. As for the remote service, we currently have no attributes other than its name and the name of its peer. However, the name is available on both sides of the connection.
3. Load-balancing policies are currently only based on names: the name of the requesting workload and the name of the remote service. This might be fine, but still requires the requesting workload to have a label with some predefined special key (currently `app`).

My plans for the short term:
1. Only accept access policies with attributes that refer to the src and dst peers or to the dst service name - reject all other policies
2. As for load-balancing policies - remain with the current scheme, but maybe use a slightly less generic label key.

For the long term we need to find a way to share the attributes of the requesting workload and the remote service between peers.

@elevran , @kfirtoledo @orozery @praveingk  - your thoughts?

kfirtoledo:
You are right about the issue of the source service In the remote peer. Butt I think is really useful that we can have control at the service level of which service can access the remote peer/service. I think also that's what @huang195 is doing.

elevran:
> 1. There is currently no way to get the attributes of the requesting workload on the remote service side. We only know the name of the requesting peer. This means ingress connection requests cannot be evaluated the same way as the corresponding egress connection requests.

I think we would want to support metadata/attribute exchange as part of the connection set-up request. For example, we can send the requesting workload attributes via JWT claims in an `Authorization` header.
There is still asymmetry as the server side attributes are not known at the client side. We could consider adding attributes to service Import (e.g., the set of labels on the k8s Service object may make sense).

> 3. Load-balancing policies are currently only based on names: the name of the requesting workload and the name of the remote service. This might be fine, but still requires the requesting workload to have a label with some predefined special key (currently `app`).

We need to have a clear definition of the attribute model and then make sure we have them present. A single label with a name is neither flexible nor sufficient, IMO. Attributes can include a service name, but likely more than that (e.g., peer attributes, workload attributes, ...)

> My plans for the short term:
> 
> 1. Only accept access policies with attributes that refer to the src and dst peers or to the dst service name - reject all other policies
[...]

PR #277, opened 2024-01-22
Preparing for workload attributes support

zivnevo:
In this PR:
* Moving structs and consts from `pkg/controlplane/eventmanager/events.go` to `policyengine`. Most of them are now in `pkg/policyengine/policytypes/connection_request.go`
* Connection request now requires providing all attributes of the source workload.
* Changed interface of `AddExport()` - will one day return a list of peers to which export is allowed (currently always returns all peers, and the returned value is currently being ignored by the controlplane.
* Removed some unused code/consts/types

References #17


pkg/policyengine/PolicyDispatcher_test.go
```diff
@@ -97,14 +96,16 @@ func TestIncomingConnectionRequests(t *testing.T) {
 	policy2.To = []policytypes.WorkloadSetOrSelector{{WorkloadSelector: &selectAllSelector}}
 	addPolicy(t, &policy2, ph)
 
-	requestAttr := event.ConnectionRequestAttr{SrcService: svcName, Direction: event.Incoming}
-	connReqResp, err := ph.AuthorizeAndRouteConnection(&requestAttr)
-	require.Equal(t, event.Allow, connReqResp.Action)
+	srcAttrs := policytypes.WorkloadAttrs{policyengine.ServiceNameLabel: svcName}
```
elevran, line 99: how/where are these retrieved in non-test code? We need to implement some way to get attributes beyond just labels being used now.

Also (and unrelated to this specific location/file), while it may be ok to "send" all properties as `map[string]string` on the wire, I wonder if we want to make them explicit by defining the individual attributes on some enclosing struct passed between control plane and the policy engine (ie., explicitly have `Image`, `Labels` etc. when collected before it is exchanged on the wire with the peer)?
It may help in reasoning in the code, about what is needed and being examined for a policy decision, easier.

pkg/policyengine/policytypes/connection_request.go
```diff
@@ -0,0 +1,36 @@
...
+
+type Direction int
+
+const (
+	Incoming Direction = iota
+	Outgoing
+)
+
+type ConnectionRequest struct {
+	SrcPeer string
```
elevran, line 24: See earlier discussion regarding use of explicit `Source` and `Destination` structs to store the relevant attributes and making the attributes explicit in the communication between control plane and policy engine (instead of map[string]string as used for current state where attributes == Labels.

pkg/policyengine/policytypes/connection_request.go
```diff
@@ -0,0 +1,36 @@
...
+	DstPeer string // ignored on an outgoing connection request
```
elevran, line 25: nit: unclear then when the destination peer is used if ignored on the outgoing request...

Perhaps a more detailed comment (per each exported field) would help?

pkg/policyengine/policytypes/connection_request.go
```diff
@@ -0,0 +1,36 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package policytypes
+
+type Direction int
```
elevran, line 16: missing export comments in multiple places

elevran (changes requested):

pkg/policyengine/policytypes/connection_request.go
zivnevo: Added

pkg/policyengine/policytypes/connection_request.go
zivnevo: Indeed, `DstPeer` wasn't used, and was removed.
`SrcPeer` is now encoded in the source-workload attributes, so removed it too.

pkg/controlplane/authz.go
```diff
@@ -80,23 +81,23 @@ func (cp *Instance) AuthorizeEgress(req *EgressAuthorizationRequest) (*EgressAut
 		return nil, fmt.Errorf("no bindings found for import '%s'", req.Import)
 	}
 
-	connReq := eventmanager.ConnectionRequestAttr{DstService: req.Import, Direction: eventmanager.Outgoing}
+	connReq := policytypes.ConnectionRequest{DstSvcName: req.Import, Direction: policytypes.Outgoing}
 	srcLabels := cp.platform.GetLabelsFromIP(req.IP)
 	if src, ok := srcLabels["app"]; ok { // TODO: Add support for labels other than just the "app" key.
 		cp.logger.Infof("Received egress authorization srcLabels[app]: %v.", srcLabels["app"])
-		connReq.SrcService = src
+		connReq.SrcWorkloadAttrs = policytypes.WorkloadAttrs{policyengine.ServiceNameLabel: src}
```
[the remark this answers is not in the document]
zivnevo: `SrcWorkloadAttrs` is now the only information on the source, as `SrcPeer` was removed. If needed, we can encapsulate more logic into `policytypes.WorkloadAttrs`

pkg/policyengine/PolicyDispatcher_test.go
zivnevo: Yes, the current PR is only preparing for the real gathering and transfer of attributes

elevran (approved):
Please open issues/follow up PRs for items left open in this review

PR #278, opened 2024-01-23
Makefile: don't use source command

orozery:
Before this PR, I was not able to run `make prereqs`:
```
oro@oro:~/go/src/github.com/clusterlink-net/clusterlink$ make prereqs
installing dev tooling...
make: source: No such file or directory
make: *** [Makefile:24: prereqs] Error 127
```

See here also:
https://stackoverflow.com/questions/44052093/makefile-with-source-get-error-no-such-file-or-directory

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+SHELL=/bin/bash
+
 IMAGE_VERSION ?= latest
 IMAGE_ORG ?= clusterlink-net
 IMAGE_BASE ?= ghcr.io/$(IMAGE_ORG)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -21,10 +21,10 @@ dist: ; $(info creating dist directory...)
 .PHONY: prereqs prereqs-force
 
 prereqs: ; $(info installing dev tooling...) 
-	@source ./hack/install-devtools.sh
+	./hack/install-devtools.sh
```
elevran, line 24: The use of `source` was needed in order to affect the environment variables (specifically, ensure $GOPATH/bin is in the PATH).
Without `source`, the command is invoked in a child shell and any environment change is reverted upon exit.
So the change, as is, would break that.

According to the SO question you linked, alternatives might be to add `SHELL=/bin/bash` to the Makefile (`source` will be found then since it's a bash built-in) or use dot (`.`) as the POSIX way of sourcing (never tried this in makefile, quick check in a shell shows it works):
```make
   @. ./hack/install-devtools.sh
```

elevran (changes requested):

elevran (approved):

PR #287, opened 2024-01-28
Design for CRD based management

elevran:
Detailed design document

### DIFF
--- a/design-proposals/crd-based-management.md
+++ b/design-proposals/crd-based-management.md
@@ -1,0 +1,417 @@
+# Custom Resource Based Management of ClusterLink
+
+**Authors**: @elevran, @oro
+
+**Begin Design Discussion**: 2023-11-22
+
+**(optional) Status:** draft
+
+## Summary/Abstract
+
+This design proposal explores the design space for managing ClusterLink through
+ Kubernetes (k8s) Custom Resource Definitions (CRDs). It aims to replace the
+ current API and CLI with k8s native management philosophy (e.g., asynchronous
+ reconciliation of actual and desired states) and tooling (e.g., `kubectl`, RBAC).
+ This document focuses on the management of a single cluster and ignores aspects
+ relating to centralized or distributed management of a ClusterLink system across
+ multiple clusters. In the context of the current design, any information
+ needed by a cluster regarding remote peers, such as their gateway identities and
+ endpoints, services they expose, etc., is assumed to be exchanged out of band.
+
+## Background
+
+### Motivation and problem space
+
+Currently ClusterLink exposes a RESTful API for its users. A convenience CLI (`gwctl`)
+ is provided for easier API access. The API/CLI is based on Created/Read/Update/Delete
+ (CRUD) of the following objects:
+
+- The `local ClusterLink peer` (albeit this not made explicit today);
+- Remote `Peer` clusters the local peer can interact with;
+- Service `Export`s the local peer makes available for consumption to remote clusters;
+- Service `Import`s made available to workloads by the local peer (note: the current API
+  also includes `Binding` objects, associating an `Import` with `Peer` clusters. This is
+  no longer deemed required).
+- `Policy` objects governing aspects of the cross cluster communication.
+
+An additional CLI (`cl-adm`) is provided to ease the bootstrap and deployment of ClusterLink
+ in a k8s cluster. The `cl-adm` utility creates a set of (mostly k8s configuration) files on
+ the local file systems which can be used to bootstrap the ClusterLink installation. This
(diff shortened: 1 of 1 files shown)


design-proposals/crd-based-management.md
```diff
@@ -0,0 +1,417 @@
...
+ the desired state asynchronously, over time. Thus, users should not expect immediate
+ feedback and state change. Other relevant practices should be followed as much as possible
+ (e.g., addition of Finalizers where appropriate, reconciliation retrieving up to date
+ information instead of using internal caches for actual state, etc,).
+
+Objects should use Conditions for tracking their Status. The following conventions for
+ status conditions are suggested. Controllers should use the following conditions types
+ to signal whether a resource has been reconciled, or if it encountered any problems:
+
+- `Reconciling`: Indicates that the resource actual state does not yet match its desired
```
kfirtoledo, line 167: I am not sure if reconciling is the best word in deployment they use progressing

design-proposals/crd-based-management.md
```diff
@@ -0,0 +1,417 @@
...
+}
+
+type ExportStatus struct {
+  ObservedGeneration int64
+  Conditions[] metav1.Condition
+  // TODO: are any additional fields needed? For example, should we reflect the actual port here?
+}
+```
+
+TODO: how and where are Service attributes defined? Both import and export are set
```
kfirtoledo, line 254: Why do peers have an attribute field while exports and imports do not?

design-proposals/crd-based-management.md
```diff
@@ -0,0 +1,417 @@
...
+
+## User Stories
+
+### Users
+
+We envisage multiple user roles involved in the management of a ClusterLink deployment.
+
+- **Fabric administrator** has control over the entire ClusterLink fabric. They are responsible
+ for creating the fabric's root of trust, controlling which peers are part of the fabric, etc.
+ This role is mostly not in scope for this design document.
```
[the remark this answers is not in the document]
elevran: Agree. This is the minimal work needed to define their existence and I think it is good enough (informal but conveys the intent). 
Once we have more experience operating the CL system, we can revisit and formalize.

design-proposals/crd-based-management.md
elevran: We want to capture Service attributes on the Import/Export. The issue is that these are set by namespace users and Service attributes seem to be more sensitive than to allow arbitrary set by any user (they affect the policies).
Import seems the least likely object for it, Exports are better, but having these in the privileged namespace would be best, IMHO

zivnevo (approved):
LGTM.
I opened a [discussion](https://github.com/clusterlink-net/clusterlink/discussions/296) on setting load-balancing policies.

design-proposals/crd-based-management.md
```diff
@@ -0,0 +1,417 @@
...
+- **Site administrator** has control over a single cluster in the fabric and manages the
+ installation and operation of the local ClusterLink deployment. Site administrators can
+ add remote Peers to their cluster and may define admin level network policies.
+- **Application owners and developers** manage Services and their access policies. They
```
praveingk, line 91: Do we support scenarios where fabric admin, site admin and application owners belong to the same entity? In this case, Clusterlink control plane and dataplane would be deployed on a single namespace which the user has access to. Ofcourse we wouldn't have CRD for that scenario. But, I am just concerned if we remove the API access, how we intend to support such scenarios.

elevran: I would differentiate between two aspects:
- the CRDs are installed (one time operation) and no further admin level access is required; and
- absolutely no global cluster admin privileges.

The first is on our roadmap, and this design doc "hints" at what could be done to support it (e.g., Peer are defined in the same namespace as the CP and DP). I'm not sure how this would work with a true "no cluster privileges" mode as we need to expose the CP/DP Service outside of the cluster which (IMHO) is a privileged operation. Think of the case of multiple different namespace, each with their own deployment - how would you route from the outside to the correct CL instance? You would need to open multiple ingress points into the cluster and I'm not sure this would be considered "unprivileged". Perhaps we can explore the use case you have in mind in more details. Note that if one time admin access is reasonable (e.g., CRD and CL are deployed) all other management would be per namespace.

The second is currently out of scope, but we might be able to support it by "emulating" CRDs as ConfigMap objects (e.g., with `version`, `type`, `spec.*` and `status.*` fields. You lose the type validation provided by the API for CRDs, but the changes in the controllers may be minimized by adding a layer to convert between ConfigMap and the corresponding Go structure for the CRD and then handing the "CRD instance" for further processing.

orozery (approved):

PR #295, opened 2024-02-04
operator/controller: Add controller unit test

kfirtoledo:
Adding deployment operator unit test using envtests environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,16 +76,20 @@ GO ?= CGO_ENABLED=0 go
 # Allow setting of go build flags from the command line.
 GOFLAGS := 
 
-# Location to install dependencies to
-GOBIN ?= $(GOPATH)/bin
-
+# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
+ifeq (,$(shell go env GOBIN))
+GOBIN=$(shell go env GOPATH)/bin
+else
+GOBIN=$(shell go env GOBIN)
+endif
 
 # Controller Gen for crds
 CONTROLLER_GEN ?= $(GOBIN)/controller-gen
 CONTROLLER_TOOLS_VERSION ?= v0.13.0
 
+.PHONY: controller-gen
 controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary. If wrong version is installed, it will be overwritten.
-$(CONTROLLER_GEN):
+$(CONTROLLER_GEN): $(GOBIN)
 	test -s $(GOBIN)/controller-gen && $(GOBIN)/controller-gen --version | grep -q $(CONTROLLER_TOOLS_VERSION) || \
 	go install sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_TOOLS_VERSION)
 
@@ -104,7 +108,6 @@ build:
 	$(GO) build -o ./bin/cl-go-dataplane ./cmd/cl-go-dataplane
 	$(GO) build -o ./bin/cl-adm ./cmd/cl-adm
 	$(GO) build -o bin/manager ./cmd/cl-operator/main.go
-
 
 docker-build: build
 	docker build --progress=plain --rm --tag cl-controlplane -f ./cmd/cl-controlplane/Dockerfile .
@@ -133,8 +136,18 @@ clean-tests:
 #------------------------------------------------------
 # Run Targets
 #------------------------------------------------------
-unit-tests:
+# Envtest use for checking the deployment operator
+ENVTEST ?= $(GOBIN)/setup-envtest
+ENVTEST_VERSION ?= latest
+ENVTEST_K8S_VERSION = 1.28.0
+.PHONY: envtest
+envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
+$(ENVTEST): $(GOBIN)
+	test -s $(GOBIN)/setup-envtest || GOBIN=$(GOBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@$(ENVTEST_VERSION)
+
+unit-tests: envtest
 	@echo "Running unit tests..."
+	export KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(GOBIN) -p path)";\
 	$(GO) test -v -count=1 ./pkg/...  -json -cover | tparse --all
 
 tests-e2e-k8s:
--- a/hack/install-devtools.sh
+++ b/hack/install-devtools.sh
@@ -86,3 +86,10 @@ if [ -z "$(which tparse)" ] || [ "$1" = "--force" ]; then
   echo installing tparse "($VERSION)"
   go install "github.com/mfridman/tparse@$VERSION"
 fi
+
+#--  setup-envtest
+VERSION=latest
+if [ -z "$(which  setup-envtest)" ] || [ "$1" = "--force" ]; then
+  echo installing  setup-envtest "($VERSION)"
+  go install "sigs.k8s.io/controller-runtime/tools/setup-envtest@$VERSION"
+fi
--- a/pkg/operator/controller/instance_controller.go
+++ b/pkg/operator/controller/instance_controller.go
@@ -158,10 +158,6 @@ func (r *InstanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		", ImageTag: ", instance.Spec.ImageTag,
 	)
 
-	if instance.Spec.ContainerRegistry != "" && instance.Spec.ContainerRegistry[len(instance.Spec.ContainerRegistry)-1:] != "/" {
-		instance.Spec.ContainerRegistry += "/"
-	}
-
 	// Set Finalizer if needed
 	if err := r.createFinalizer(ctx, instance); err != nil {
 		return ctrl.Result{}, fmt.Errorf("can't create Finalizer %w", err)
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -73,16 +73,22 @@ GO ?= CGO_ENABLED=0 go
 # Allow setting of go build flags from the command line.
 GOFLAGS := 
 
-# Location to install dependencies to
-GOBIN ?= $(GOPATH)/bin
+# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
```
elevran, line 76: did the previous version (using `?=`) not work?

Makefile
```diff
@@ -130,8 +135,17 @@ clean-tests:
...
 # Run Targets
 #------------------------------------------------------
-unit-tests:
+# Envtest use for checking the deployment operator
+ENVTEST ?= $(GOBIN)/setup-envtest
+ENVTEST_K8S_VERSION = 1.28.0
+.PHONY: envtest
+envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
+$(ENVTEST): $(GOBIN)
+	test -s $(GOBIN)/setup-envtest || GOBIN=$(GOBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
```
elevran, line 144: shouldn't this be set to `@$ENVTEST_K8S_VERSION` instead of `@latest`?

pkg/operator/controller/instance_controller_test.go
```diff
@@ -0,0 +1,319 @@
...
+	rbacv1 "k8s.io/api/rbac/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/scheme"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+
+	clusterlink "github.com/clusterlink-net/clusterlink/pkg/apis/clusterlink.net/v1alpha1"
```
elevran, line 39: consider import name along the same lines as used for k8s (e.g., `clv1a1`)

pkg/operator/controller/instance_controller_test.go
```diff
@@ -0,0 +1,319 @@
...
+	"github.com/clusterlink-net/clusterlink/pkg/operator/controller"
```
elevran, line 40: qq: what was the rationale for having a subpackage under `operator`? Both operator and controller are very generic names so I'm not understanding the value add.

pkg/operator/controller/instance_controller_test.go
```diff
@@ -0,0 +1,319 @@
...
+		err := k8sManager.Start(ctx)
+		if err != nil {
+			fmt.Printf("Failed to run controller manager: %v\n", err)
+		}
+	}()
+
+	code := m.Run()
+
+	// Cancelling the context
+	cancel()
```
elevran, line 124: if `m.Run` is synchronous, whatis the purpose of calling `cancel`? Context cancelling is usually done to interrupt a blocking go routine.

pkg/operator/controller/instance_controller_test.go
```diff
@@ -0,0 +1,319 @@
...
+	os.Exit(code)
+}
+
+func TestClusterLinkController(t *testing.T) {
+	RegisterTestingT(t) // Setup gomega
+	// Define utility constants for object names and testing timeouts/durations and intervals.
+
+	cl := clusterlink.Instance{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "batch.tutorial.kubebuilder.io/v1",
```
elevran, line 141: This seems like the wrong API version

pkg/operator/controller/instance_controller_test.go
```diff
@@ -0,0 +1,319 @@
...
+	cpID := types.NamespacedName{Name: controller.ControlPlaneName, Namespace: controller.InstanceNamespace}
+	cpResource := []client.Object{
+		&appsv1.Deployment{}, &corev1.Service{}, &corev1.PersistentVolumeClaim{},
+		&rbacv1.ClusterRole{}, &rbacv1.ClusterRoleBinding{},
+	}
+	dpID := types.NamespacedName{Name: controller.DataPlaneName, Namespace: controller.InstanceNamespace}
+	dpResource := []client.Object{&appsv1.Deployment{}, &corev1.Service{}}
+	ingressID := types.NamespacedName{Name: controller.IngressName, Namespace: controller.InstanceNamespace}
+
+	t.Run("Create ClusterLink deployment", func(t *testing.T) {
```
elevran, line 159: Is the gomega style testing required? It differs from everything else we have in the code base.

pkg/operator/controller/instance_controller_test.go
```diff
@@ -0,0 +1,319 @@
...
+		// Check Dataplane fields
+		dp := &appsv1.Deployment{}
+		getResource(dpID, dp)
+		envoyImage := "ghcr.io/clusterlink-net/" + controller.DataPlaneName + ":latest"
+		Expect(dp.Spec.Template.Spec.Containers[0].Image).Should(Equal(envoyImage))
+		Expect(*dp.Spec.Replicas).Should(Equal(int32(1)))
+		Expect(dp.Spec.Template.Spec.Containers[0].Args[1]).Should(Equal("info"))
+
+		// Check ingress resource not exist
+		assert.True(t, checkResourceNotExist(ingressID, &corev1.Service{}))
```
elevran, line 204: `assert` uses testify and not gomega (e.g., `Expect().Should(...)`)

pkg/operator/controller/instance_controller_test.go
```diff
@@ -0,0 +1,319 @@
...
+
+		// Update dataplane and controlpane
+		goReplicas := 2
+		loglevel := "debug"
+		containerRegistry := "quay.com"
+		imageTag := "v1.0.1"
+		cp := &appsv1.Deployment{}
+		dp := &appsv1.Deployment{}
+
+		getResource(cpID, cp)
```
elevran, line 234: shouldn`t deletion by done by the operator in response to changes in the Spec? Why do you need to manually delete (which I would expect the controller to be notified on and recreate the resource immediately...)?

pkg/operator/controller/instance_controller_test.go
```diff
@@ -0,0 +1,319 @@
...
+		}
+
+		// Dataplane resources
+		for _, r := range dpResource {
+			checkResourceDeleted(dpID, r)
+		}
+
+		// Ingress resource
+		checkResourceDeleted(ingressID, &corev1.Service{})
+	})
```
elevran, line 278: Can you please add a test for multiple clusterlink.Instance objects and their changes? Only the first should be in effect and when deleted the next in line should be realized.

elevran (changes requested):

Makefile
kfirtoledo: It didn't work in the Github action

Makefile
kfirtoledo: the ENVTEST_K8S_VERSION (is the k8s version) that needs to match the setup-envtest.

pkg/operator/controller/instance_controller_test.go
kfirtoledo: I can change it, though maybe the operator will have more sub-packages

pkg/operator/controller/instance_controller_test.go
kfirtoledo: I used it for cleanup the context

pkg/operator/controller/instance_controller_test.go
kfirtoledo: I need it just for the "Eventually" function that uses deployed units when we create them, for all other cases I can use assert, but I thought to keep it in the same style for this specific test.

elevran: Please use a consistent style as other tests in the repo, not using Gomega style.
If you need a specific function, such as Eventually, maybe it's available in testify already or can be implemented locally?

pkg/operator/controller/instance_controller_test.go
kfirtoledo: Done in the e2e test

Makefile
kfirtoledo: I can replace it with: v0.0.0-20240215124517-56159419231e
But I prefer to use it like the Kubebuilder: https://github.com/kubernetes-sigs/kubebuilder/blob/8afeb403549fa87bf55a00e10a819a34719eec53/testdata/project-v4/Makefile#L169

kfirtoledo:
@elevran PTAL

hack/install-devtools.sh
```diff
@@ -86,3 +86,10 @@ if [ -z "$(which tparse)" ] || [ "$1" = "--force" ]; then
   echo installing tparse "($VERSION)"
   go install "github.com/mfridman/tparse@$VERSION"
 fi
+
+#--  setup-envtest
```
elevran, line 90: This should be added to the dev-container file as well (`.devcontainer/dev/post-create.sh`)

pkg/operator/controller/instance_controller.go
```diff
@@ -543,18 +541,21 @@ func (r *InstanceReconciler) deleteFinalizer(ctx context.Context, instance *clus
...
+	key := client.ObjectKeyFromObject(object)
+	err := r.Client.Create(ctx, object)
+	if err == nil {
+		r.Logger.Infof("Create resource %s Name: %s Namespace: %s", reflect.TypeOf(object), object.GetName(), object.GetNamespace())
 		return nil
-	})
+	}
+
+	if errors.IsAlreadyExists(err) {
+		// If resource already exists, update it
```
elevran, line 553:
```suggestion
	if errors.IsAlreadyExists(err) { // If resource already exists, update it
```

pkg/operator/controller/instance_controller.go
```diff
@@ -543,18 +541,21 @@ func (r *InstanceReconciler) deleteFinalizer(ctx context.Context, instance *clus
...
+		err = r.Client.Update(ctx, object)
```
elevran, line 554: is there a need to Get the resource first so that you only override specific fields (and also have the correct resource version)?
Not sure...

elevran (approved):

PR #299, opened 2024-02-06
apis: Add Export, Import, Peer, AccessPolicy CRDs

orozery:
This commit adds the above objects and their respective CRDs.


pkg/apis/clusterlink.net/v1alpha1/import.go
```diff
@@ -0,0 +1,71 @@
...
+	Spec ImportSpec `json:"spec"`
+}
+
+// ImportSource represents an addressable exported service.
+type ImportSource struct {
+	// Peer name where the exported service is defined.
+	Peer string `json:"peer"`
+	// ExportName is the name of the exported service.
+	ExportName string `json:"name"`
+	// ExportNamespace is the namespace of the exported service.
```
kfirtoledo, line 38: the namespace it requires? if is empty?

pkg/apis/clusterlink.net/v1alpha1/import.go
```diff
@@ -0,0 +1,71 @@
...
+	ExportNamespace string `json:"namespace"`
+}
+
+// ImportSpec contains all attributes of an imported service.
+type ImportSpec struct {
+	// Port of the imported service.
+	Port uint16 `json:"port"`
+	// TargetPort of the imported service.
+	TargetPort uint16 `json:"targetPort,omitempty"`
+	// Peers are the list of peers to import from.
```
kfirtoledo, line 48: Need to fix the comment

pkg/apis/clusterlink.net/v1alpha1/export.go
```diff
@@ -0,0 +1,57 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec represents the attributes of the exported service.
+	Spec ExportSpec `json:"spec,omitempty"`
+}
+
+// ExportSpec contains all attributes of an exported service.
+type ExportSpec struct {
+	// Port of the exported service.
```
kfirtoledo, line 35: What about the service?
https://github.com/clusterlink-net/clusterlink/pull/287/files#:~:text=ExportSpec%20struct%20%7B-,Service%20string,-//%20name%20of

kfirtoledo (changes requested):

pkg/apis/clusterlink.net/v1alpha1/export.go
orozery: For start I only implement support for service which equals the export name.
If we add support for custom service name, we can add that field then.

pkg/apis/clusterlink.net/v1alpha1/import.go
elevran: Since exports are created by namespace owners, they can create overlapping names (since the object's full name is <ns>/<name>). We need to know, explicitly, which export we refer to.

orozery:
@kfirtoledo please note I also added an AccessPolicy object

pkg/apis/clusterlink.net/v1alpha1/accesspolicy.go
```diff
@@ -0,0 +1,74 @@
...
+	AccessPolicyActionDeny  AccessPolicyAction = "deny"
+)
+
+// WorkloadSetOrSelectorList is a collection of WorkloadSetOrSelector objects.
+type WorkloadSetOrSelectorList []WorkloadSetOrSelector
+
+// WorkloadSetOrSelector describes a set of workloads, based on their attributes (labels)
+// Exactly one of the two fields should be non-empty.
+type WorkloadSetOrSelector struct {
+	WorkloadSets     []string              `json:"workloadSets,omitempty"`
```
kfirtoledo, line 49: Can you add a description to the fields?

pkg/apis/clusterlink.net/v1alpha1/accesspolicy.go
```diff
@@ -0,0 +1,74 @@
...
+	WorkloadSelector *metav1.LabelSelector `json:"workloadSelector,omitempty"`
+}
+
+// AccessPolicySpec contains all attributes of an access policy.
+type AccessPolicySpec struct {
+	Privileged bool                      `json:"privileged"`
```
kfirtoledo, line 55: Same

kfirtoledo (changes requested):

kfirtoledo (approved):

PR #314, opened 2024-02-14
tests/e2e: Add support to create cluster with operator

kfirtoledo:
Add options to create Clustrlink using the operator in the e2e tests.

### DIFF
--- a/pkg/operator/controller/instance_controller.go
+++ b/pkg/operator/controller/instance_controller.go
@@ -47,7 +47,7 @@ const (
 	DataPlaneName     = "cl-dataplane"
 	GoDataPlaneName   = "cl-go-dataplane"
 	IngressName       = "clusterlink"
-	OperatorNameSpace = "clusterlink-operator" // TODO -should be removed
+	OperatorNamespace = "clusterlink-operator"
 	InstanceNamespace = "clusterlink-system"
 	FinalizerName     = "instance.clusterlink.net/finalizer"
 
@@ -863,7 +863,7 @@ func (r *InstanceReconciler) updateCondition(conditions map[string]metav1.Condit
 func (r *InstanceReconciler) triggerAnotherInstance(ctx context.Context, name, namespace string) error {
 	var earliestInstance *clusterlink.Instance
 	instanceList := &clusterlink.InstanceList{}
-	if err := r.List(ctx, instanceList, client.InNamespace(OperatorNameSpace)); err != nil {
+	if err := r.List(ctx, instanceList, client.InNamespace(OperatorNamespace)); err != nil {
 		return err
 	}
 
--- a/tests/e2e/k8s/suite.go
+++ b/tests/e2e/k8s/suite.go
@@ -32,7 +32,7 @@ const (
 	clusterCount = 3
 )
 
-var images = [...]string{"cl-controlplane", "cl-dataplane", "cl-go-dataplane"}
+var images = [...]string{"cl-controlplane", "cl-dataplane", "cl-go-dataplane", "cl-operator"}
 
 var iperf3Service = util.Service{
 	Name:      "iperf3-server",
@@ -98,8 +98,21 @@ func (s *TestSuite) SetupSuite() {
 		if err != nil {
 			s.T().Fatal(fmt.Errorf("cannot create iperf3-server service: %w", err))
 		}
-	}
 
+		// create the project operator.
+		if err := s.clusters[i].CreateFromPath("./../../../config/operator/manager/"); err != nil {
+			s.T().Fatal(fmt.Errorf("cannot create k8s operator deployment files: %w", err))
+		}
+
+		if err := s.clusters[i].CreateFromPath("./../../../config/operator/rbac/"); err != nil {
+			s.T().Fatal(fmt.Errorf("cannot create k8s operator rbac files: %w", err))
+		}
+
+		// create the project CRDs.
+		if err := s.clusters[i].CreateFromPath("./../../../config/operator/crds/"); err != nil {
+			s.T().Fatal(fmt.Errorf("cannot create project CRDs: %w", err))
+		}
+	}
 	// wait for fabric
 	if err := fabric.Wait(); err != nil {
 		s.T().Fatal(err)
--- a/tests/e2e/k8s/util/fabric.go
+++ b/tests/e2e/k8s/util/fabric.go
@@ -20,11 +20,16 @@ import (
 	"fmt"
 	"time"
 
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/e2e-framework/klient/wait"
+	"sigs.k8s.io/e2e-framework/klient/wait/conditions"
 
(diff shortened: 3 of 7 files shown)


tests/e2e/k8s/util/kind.go
```diff
@@ -151,6 +154,11 @@ func (c *KindCluster) initializeClients() error {
 		return fmt.Errorf("unable to initialize k8s clientset: %w", err)
 	}
 
+	// Add instance CRD to scheme.
```
orozery, line 157:
```suggestion
	// Add clusterlink CRDs to scheme.
```

Also, my personal reference is to rename an import (i.e. `v1alpha1` to `clusterlink`) only when package names collide. But your call.

tests/e2e/k8s/util/kind.go
```diff
@@ -361,6 +369,14 @@ func (c *KindCluster) CreateFromYAML(yaml, namespace string) error {
 		decoder.MutateNamespace(namespace))
 }
 
+// CreateFromFolder creates k8s objects from a yaml in a folder.
+func (c *KindCluster) CreateFromFolder(folder string) error {
```
orozery, line 373: I think folder is a Windows term. Perhaps rename to Directory?

tests/e2e/k8s/util/kind.go
```diff
@@ -361,6 +369,14 @@ func (c *KindCluster) CreateFromYAML(yaml, namespace string) error {
...
+	pattern := "*"
+	fs := os.DirFS(folder)
+	return decoder.DecodeEachFile(context.TODO(), fs, pattern,
```
orozery, line 376:
```suggestion
	return decoder.DecodeEachFile(context.Background(), fs, pattern,
```

tests/e2e/k8s/util/kind.go
```diff
@@ -361,6 +369,14 @@ func (c *KindCluster) CreateFromYAML(yaml, namespace string) error {
 		decoder.MutateNamespace(namespace))
 }
 
+// CreateFromFolder creates k8s objects from a yaml in a folder.
+func (c *KindCluster) CreateFromFolder(folder string) error {
+	pattern := "*"
+	fs := os.DirFS(folder)
```
orozery, line 375: I would remove this lines, and embed the values directly to the function call below.

tests/e2e/k8s/util/kind.go
```diff
@@ -470,6 +486,19 @@ func (c *KindCluster) ScaleDeployment(name, namespace string, replicas int32) er
...
+		return err
+	}
+	return c.CreateFromFolder("./../../../config/operator/rbac/")
+}
+
+// CreateCRDs creates the CRDs from folder config.
+func (c *KindCluster) CreateCRDs() error {
+	return c.CreateFromFolder("./../../../config/operator/crds/")
+}
+
```
orozery, line 501: I'm trying to keep this file to be general kind-utils, and not clusterlink specific.
I would remove those functions, and inline their code in the caller.

tests/e2e/k8s/util/fabric.go
```diff
@@ -146,23 +153,75 @@ func (f *Fabric) SwitchToNewNamespace(name string, appendName bool) error {
...
+
+// deployUsingOperator deploys ClusterLink using operator.
+func (f *Fabric) deployUsingOperator(target *peer, cfg *PeerConfig) error {
+	// Create ClusterLink instance
+	instance, err := f.generateClusterlinkInstance(target, cfg)
+	if err != nil {
+		return fmt.Errorf("cannot generate ClusterLink instance: %w", err)
+	}
+
+	if err := target.cluster.CreateFromYAML(instance, controller.OperatorNameSpace); err != nil {
```
orozery, line 166: Just noticed this const. NameSpace -> Namespace

orozery (changes requested):

tests/e2e/k8s/util/kind.go
elevran: Or `path`.
Would it also make sense to validate that the parameter is indeed a directory and not a file or let `DirFS` fail?

kfirtoledo:
@orozery PTAL (After the approval I will sync it).

pkg/operator/controller/instance_controller.go
```diff
@@ -47,7 +47,7 @@ const (
 	DataPlaneName     = "cl-dataplane"
 	GoDataPlaneName   = "cl-go-dataplane"
 	IngressName       = "cl-external-svc"
-	OperatorNameSpace = "clusterlink-operator" // TODO -should be removed
+	OperatorNamespace = "clusterlink-operator"
```
orozery, line 50: Why is the default operator namespace different than the instance namespace?

tests/e2e/k8s/suite.go
```diff
@@ -98,8 +98,20 @@ func (s *TestSuite) SetupSuite() {
 		if err != nil {
 			s.T().Fatal(fmt.Errorf("cannot create iperf3-server service: %w", err))
 		}
-	}
+		// create the project operator.
```
orozery, line 101:
```suggestion

		// create the project operator.
```

tests/e2e/k8s/util/fabric.go
```diff
@@ -20,11 +20,16 @@ import (
...
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/clusterlink-net/clusterlink/pkg/bootstrap"
 	"github.com/clusterlink-net/clusterlink/pkg/bootstrap/platform"
 	"github.com/clusterlink-net/clusterlink/pkg/client"
+	"github.com/clusterlink-net/clusterlink/pkg/operator/controller"
+	"sigs.k8s.io/e2e-framework/klient/wait"
+	"sigs.k8s.io/e2e-framework/klient/wait/conditions"
```
orozery, line 32: Move these to the upper import group.

tests/e2e/k8s/util/fabric.go
```diff
@@ -37,6 +42,8 @@ type PeerConfig struct {
 	ControlplanePersistency bool
 	// ExpectLargeDataplaneTraffic hints that a large amount of dataplane traffic is expected.
 	ExpectLargeDataplaneTraffic bool
+	// DeployWithOperator deploy cluster using the operator.
```
orozery, line 45:
```suggestion
	// DeployWithOperator deploys clusterlink using an operator.
```

tests/e2e/k8s/util/kind.go
```diff
@@ -17,11 +17,13 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
 	"reflect"
 	"strings"
 	"sync"
 	"time"
 
+	clusterlink "github.com/clusterlink-net/clusterlink/pkg/apis/clusterlink.net/v1alpha1"
```
orozery, line 26: move this import to its own import group downwards.

orozery (changes requested):

pkg/operator/controller/instance_controller.go
kfirtoledo: Usually, the operator runs in a different namespace with more privileges (like in openshift there is an openshift-operator namespace)

orozery (approved):

PR #315, opened 2024-02-14
operator/controller: Handle more than one instance in the same namespace 

kfirtoledo:
Small fixes:
1. Handle more than one instance in the same namespace 
2. Fix IngressName to cl-ingress

### DIFF
--- a/config/operator/crds/clusterlink.net_instances.yaml
+++ b/config/operator/crds/clusterlink.net_instances.yaml
@@ -78,7 +78,7 @@ spec:
                     default: none
                     description: 'Type represents the type of service used to expose
                       the ClusterLink deployment. Supported values: "LoadBalancer","NodePort",
-                      "none". The service name will be "cl-external-svc".'
+                      "none". The service name will be "clusterlink".'
                     enum:
                     - none
                     - LoadBalancer
--- a/pkg/apis/clusterlink.net/v1alpha1/instance_types.go
+++ b/pkg/apis/clusterlink.net/v1alpha1/instance_types.go
@@ -96,7 +96,7 @@ type IngressSpec struct {
 	// +kubebuilder:default=none
 	// Type represents the type of service used to expose the ClusterLink deployment.
 	// Supported values: "LoadBalancer","NodePort", "none".
-	// The service name will be "cl-external-svc".
+	// The service name will be "clusterlink".
 	Type IngressType `json:"type,omitempty"`
 
 	// Port represents the port number of the external service.
--- a/pkg/operator/controller/instance_controller.go
+++ b/pkg/operator/controller/instance_controller.go
@@ -46,7 +46,7 @@ const (
 	ControlPlaneName  = "cl-controlplane"
 	DataPlaneName     = "cl-dataplane"
 	GoDataPlaneName   = "cl-go-dataplane"
-	IngressName       = "cl-external-svc"
+	IngressName       = "clusterlink"
 	OperatorNameSpace = "clusterlink-operator" // TODO -should be removed
 	InstanceNamespace = "clusterlink-system"
 	FinalizerName     = "instance.clusterlink.net/finalizer"
@@ -59,9 +59,9 @@ const (
 // InstanceReconciler reconciles a ClusterLink instance object.
 type InstanceReconciler struct {
 	client.Client
-	Scheme        *runtime.Scheme
-	Logger        *logrus.Entry
-	InstancesMeta map[string]string
+	Scheme    *runtime.Scheme
+	Logger    *logrus.Entry
+	Instances map[string]string
 }
 
 // +kubebuilder:rbac:groups=clusterlink.net,resources=instances,verbs=list;get;watch;update;patch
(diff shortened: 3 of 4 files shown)


config/operator/crds/clusterlink.net_instances.yaml
```diff
@@ -78,7 +78,7 @@ spec:
                     default: none
                     description: 'Type represents the type of service used to expose
                       the ClusterLink deployment. Supported values: "LoadBalancer","NodePort",
-                      "none". The service name will be "cl-external-svc".'
+                      "none". The service name will be "cl-ingress".'
```
elevran, line 81:
```suggestion
                      "none". The service name will be "clusterlink".'
```

pkg/apis/clusterlink.net/v1alpha1/instance_types.go
```diff
@@ -96,7 +96,7 @@ type IngressSpec struct {
 	// +kubebuilder:default=none
 	// Type represents the type of service used to expose the ClusterLink deployment.
 	// Supported values: "LoadBalancer","NodePort", "none".
-	// The service name will be "cl-external-svc".
+	// The service name will be "cl-ingress".
```
elevran, line 99: I think the addition of `-ingress` is a somewhat redundant (somewhat akin to as adding `-ns` to namespace names)
```suggestion
	// The service name will be "clusterlink".
```

pkg/operator/controller/instance_controller.go
```diff
@@ -46,7 +46,7 @@ const (
 	ControlPlaneName  = "cl-controlplane"
 	DataPlaneName     = "cl-dataplane"
 	GoDataPlaneName   = "cl-go-dataplane"
-	IngressName       = "cl-external-svc"
+	IngressName       = "cl-ingress"
```
elevran, line 49:
```suggestion
	IngressName       = "clusterlink"
```

pkg/operator/controller/instance_controller.go
```diff
@@ -116,14 +116,16 @@ func (r *InstanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	}
 
 	// Check one clusterlink per namespace
-	if name, exist := r.InstancesMeta[instance.Namespace]; exist {
+	if name, exist := r.InstancesMeta[instance.Spec.Namespace]; exist {
 		if instance.Name != name {
-			err := fmt.Errorf("instance %s in Namespace %s is  already exist- Only one instance are permit in  a Namespace",
-				name, instance.Namespace)
+			r.Logger.Errorf(
+				"Can't create instance %s in namespace %s, instance %s  is already exist - Only one instance are permit in a Namespace",
```
elevran, line 122:
```suggestion
				"Can't create instance %s in namespace %s, instance %s already exists and only one instance is permitted in a Namespace",
```

pkg/operator/controller/instance_controller.go
```diff
@@ -116,14 +116,16 @@ func (r *InstanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	}
 
 	// Check one clusterlink per namespace
-	if name, exist := r.InstancesMeta[instance.Namespace]; exist {
+	if name, exist := r.InstancesMeta[instance.Spec.Namespace]; exist {
```
elevran, line 119: suggestion `Instances`, without the `Meta`?

pkg/operator/controller/instance_controller.go
```diff
@@ -134,6 +136,11 @@ func (r *InstanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		if err := r.deleteFinalizer(ctx, instance); err != nil {
 			return ctrl.Result{}, err
 		}
+
+		delete(r.InstancesMeta, instance.Spec.Namespace)
+		if err := r.triggerAnotherInstance(ctx, instance.Name, instance.Spec.Namespace); err != nil {
```
elevran, line 141: shouldn't this trigger without `instance.Name` which was just deleted? Isn't it enough to just call with the namespace?
Or maybe the instance with `Name` has not been deleted yet and thus should be ignored in the retrigger?

pkg/operator/controller/instance_controller.go
```diff
@@ -822,6 +857,25 @@ func (r *InstanceReconciler) updateCondition(conditions map[string]metav1.Condit
 	return update
 }
 
+// triggerAnotherInstance checks if another CRD instance exists and triggers it.
+func (r *InstanceReconciler) triggerAnotherInstance(ctx context.Context, name, namespace string) error {
+	List := &clusterlink.InstanceList{}
```
elevran, line 862: Use a lower case `list` as there is no reason to suggest an exported symbol.

pkg/operator/controller/instance_controller.go
```diff
@@ -822,6 +857,25 @@ func (r *InstanceReconciler) updateCondition(conditions map[string]metav1.Condit
...
+	if err := r.List(ctx, List); err != nil {
```
elevran, line 863: Why not call `List` on the current namespace only? Is there a way to set the List to do that?

pkg/operator/controller/instance_controller.go
```diff
@@ -822,6 +857,25 @@ func (r *InstanceReconciler) updateCondition(conditions map[string]metav1.Condit
...
+		return err
+	}
+	for i := range List.Items {
+		if List.Items[i].Spec.Namespace == namespace && List.Items[i].Name != name { // Checks for new instnace in the same namespace.
```
elevran, line 867: This will pick an arbitrary CRD object in the namespace - the first that was returned enumerating the list. I would suggest doing a fixed selection (e.g., pick the earliest instance by timestamp)

elevran (changes requested):

pkg/operator/controller/instance_controller.go
kfirtoledo: Exactly, is not been deleted yet. (see comment in line 867).

pkg/operator/controller/instance_controller.go
kfirtoledo: It doesn't matter. All the items are in the operator namespace.

elevran (approved):

pkg/operator/controller/instance_controller.go
elevran: True for now. But from k8s API point of view you're giving it a constraint to work with an optimize the listing so it does not need to consider objects in other namespaces.
In addition, constraining to the current workspace would work correctly even with per-namespace deployment (only look for replacement in the same namespace as the one deleted - RBAC in this mode would be extra protection since you wont have a way to list outside the current namespace, but I'm not sure if you'll get an access error or just the instances in namespaces you're allowed to access)

PR #325, opened 2024-02-21
controlplane: Introduce control package

orozery:
This PRadds a control package to the controlplane.
This packages is responsible for handling centralized control operations, including imported service creation and port allocation.


pkg/platform/k8s/platform.go
```diff
@@ -19,98 +19,17 @@ import (
 	logrusr "github.com/bombsimon/logrusr/v4"
```
kfirtoledo, line 19: Do we need this file?

pkg/controlplane/control/manager.go
```diff
@@ -0,0 +1,204 @@
...
+		ctx,
+		&v1.Service{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      imp.Name,
+				Namespace: imp.Namespace,
+			},
+		})
+}
+
+func serviceChanged(svc1, svc2 *v1.Service) bool {
```
kfirtoledo, line 169: there isn't a better way to do it with deep copy?

kfirtoledo (approved):
LGTM, small comments

pkg/platform/k8s/platform.go
orozery: Currently yes because of `GetLabelsFromIP`.
This will be moved in a PR I'm just about to open.
So eventually yes - this file will be removed.

pkg/controlplane/control/manager.go
orozery: hmmm I searched google at the time and did not see anything that seems like a perfect solution.
Specifically came across this thread here:
https://github.com/kubernetes-sigs/kubebuilder/issues/592

But I'm open for new insights :)

orozery: My fear is that the `ServiceSpec` returned from the API server can include some kind of auto-filled values which are not present on the `ServiceSpec` we build for creation.
@elevran your thoughts? Do you think it's safe to use `reflect.DeepEqual(oldService.Spec, newService.Spec)` ?

elevran: based on the issue you posted, it isn't.
If the auto-filled fields are only updated once then a second call (e.g., status update) would be a nil-op.
However, your code would still need to check the relevant fields on the updates to see if it needs to reconcile or not, so might as well do it as you did here.

PR #328, opened 2024-02-21
controlplane: Introduce authz package

orozery:
This comPRit adds an authz package to the controlplane.
This package is responsible for authorizing dataplane connections (either ingress or egress).


pkg/platform/k8s/pod_reconciler.go
```diff
@@ -1,128 +0,0 @@
...
-	for _, ip := range pod.Status.PodIPs {
-		// ignoring host-networked Pod IPs
-		if ip.IP != pod.Status.HostIP {
-			r.ipToPod[ip.IP] = podID
-		}
-	}
-}
-
-// GetLabelsFromIP returns the labels associated with Pod with the specified IP address.
-func (r *PodReconciler) GetLabelsFromIP(ip string) map[string]string {
```
kfirtoledo, line 99: Where is the replacement for this function?

orozery: https://github.com/orozery/clusterlink/blob/authz/pkg/controlplane/authz/manager.go#L242

kfirtoledo (approved):

PR #347, opened 2024-02-25
controlplane/control: Support port updates

orozery:
This PR adds support for updating import target ports.
It also adds support for automatic port allocation in CRD mode.

### DIFF
--- a/pkg/controlplane/control/manager.go
+++ b/pkg/controlplane/control/manager.go
@@ -36,8 +36,9 @@ import (
 // This includes target port generation for imported services, as well as
 // k8s service creation per imported service.
 type Manager struct {
-	client client.Client
-	ports  *portManager
+	client  client.Client
+	crdMode bool
+	ports   *portManager
 
 	logger *logrus.Entry
 }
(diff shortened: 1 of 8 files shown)


pkg/controlplane/control/manager.go
```diff
@@ -130,26 +123,56 @@ func (m *Manager) AddImport(ctx context.Context, imp *v1alpha1.Import) error {
...
+		if m.crdMode {
+			if err := m.client.Update(ctx, imp); err != nil {
+				m.ports.Release(port)
+				return err
+			}
+		}
+	}
+
+	if create {
+		err = m.client.Create(ctx, newService)
```
kfirtoledo, line 166: Why it can't be in line 139 + return

pkg/controlplane/control/manager.go
```diff
@@ -130,26 +123,56 @@ func (m *Manager) AddImport(ctx context.Context, imp *v1alpha1.Import) error {
...
+			return err
 		}
 
-		return err
+		create = true
 	}
 
-	if serviceChanged(&oldService, newService) {
-		return m.client.Update(ctx, newService)
+	if !create && len(oldService.Spec.Ports) == 1 && imp.Spec.TargetPort == 0 {
```
kfirtoledo, line 142: Can you write a comment that describes the logic if you decide if you need to allocate a new port it is not really clear

kfirtoledo (changes requested):

pkg/controlplane/control/manager.go
orozery: because then you have to copy code from lines 153-163 + 171-173

pkg/controlplane/control/manager.go
```diff
@@ -130,26 +123,56 @@ func (m *Manager) AddImport(ctx context.Context, imp *v1alpha1.Import) error {
...
+		imp.Spec.TargetPort = uint16(oldService.Spec.Ports[0].TargetPort.IntVal)
 	}
 
-	return nil
+	newPort := imp.Spec.TargetPort == 0
+
+	port, err := m.ports.Lease(imp.Name, imp.Spec.TargetPort)
```
elevran, line 148: Can you have the same import name in different namespaces? If so, they'd be treated as the same and it would be better to use a NamespacedName for the leasing inport

Issue #218, opened 2023-11-27
Go dataplane does not support xDS update/delete


orozery:
Currently, go dataplane does not support deletion of listeners/clusters.
Also, it does not support updating listeners:

```
// AddListener adds a listener to the map
func (d *Dataplane) AddListener(listener *listener.Listener) {
	listenerName := strings.TrimPrefix(listener.Name, api.ImportListenerPrefix)
	if _, ok := d.listeners[listenerName]; ok {
		return
	}
```

PR #360, opened 2024-02-29
Support remove/update operations in go dataplane

praveingk:


pkg/dataplane/server/dataplane.go
```diff
@@ -80,6 +95,17 @@ func (d *Dataplane) AddListener(ln *listener.Listener) {
 	}()
 }
 
+// RemoveListener removes a listener.
+func (d *Dataplane) RemoveListener(name string) {
```
orozery, line 99: This is a bit confusing:
You have:
- `CreateListener` - creates a listening socket
- `AddListener` - bookkeeping + call `CreateListener` in a new go-routine
- `DeleteListener` - async close of a listening socket
- `RemoveListener` - bookkeeping + call do `DeleteListener`

Maybe you can merge each matching two?

pkg/dataplane/server/dataplane.go
```diff
@@ -61,16 +61,31 @@ func (d *Dataplane) GetClusterHost(name string) (string, error) {
...
 
-// AddCluster adds a cluster to the map.
+// AddCluster adds/updates a cluster to the map.
 func (d *Dataplane) AddCluster(c *cluster.Cluster) {
 	d.clusters[c.Name] = c
 }
 
+// RemoveCluster adds a cluster to the map.
+func (d *Dataplane) RemoveCluster(name string) {
+	delete(d.clusters, name)
```
orozery, line 71: I just noticed that you need to add a lock to guard d.clusters (in all places where it is accessed).

orozery (changes requested):
Can you search the code for "https://github.com/clusterlink-net/clusterlink/issues/218" and handle the 2 TODOs that will enable e2e testing?

pkg/dataplane/server/dataplane.go
```diff
@@ -40,6 +41,8 @@ type Dataplane struct {
 	clusters           map[string]*cluster.Cluster
 	listeners          map[string]*listener.Listener
 	listenerEnd        map[string]chan bool
+	clustersLock       sync.Mutex
+	listenerLock       sync.Mutex
```
orozery, line 45: I don't think this is needed as I don't see any concurrent access to the listeners map.

pkg/dataplane/server/dataplane.go
```diff
@@ -40,6 +41,8 @@ type Dataplane struct {
 	clusters           map[string]*cluster.Cluster
 	listeners          map[string]*listener.Listener
 	listenerEnd        map[string]chan bool
+	clustersLock       sync.Mutex
```
orozery, line 44: I think you're missing taking a lock on `GetClusterTarget` and `GetClusterHost`.

pkg/dataplane/server/dataplane.go
```diff
@@ -61,25 +64,59 @@ func (d *Dataplane) GetClusterHost(name string) (string, error) {
...
+}
+
+// RemoveCluster adds a cluster to the map.
+func (d *Dataplane) RemoveCluster(name string) {
+	d.clustersLock.Lock()
+	delete(d.clusters, name)
+	d.clustersLock.Unlock()
+}
+
+// GetClusters returns the listeners map.
```
orozery, line 81:
```suggestion
// GetClusters returns the clusters map.
```

orozery (changes requested):

pkg/dataplane/server/dataplane.go
praveingk: Done

pkg/dataplane/server/dataplane.go
praveingk: Listener is also added and removed similar to clusters.

pkg/dataplane/server/dataplane.go
elevran: you could potentially encapsulate the map+RWLock (for clusters and listeners, separately) and then add the Update/Delete methods on the new type with internal locking. For Getting the map, you may want to create a slice copy of the names and return that to the caller or handle the clearing internally via a method

pkg/dataplane/server/dataplane.go
```diff
@@ -54,24 +59,50 @@ func (d *Dataplane) GetClusterTarget(name string) (string, error) {
...
+		if ln.Address.GetSocketAddress().GetAddress() == le.Address.GetSocketAddress().GetAddress() &&
+			ln.Address.GetSocketAddress().GetPortValue() == le.Address.GetSocketAddress().GetPortValue() {
+			d.listenerLock.RUnlock()
+			return
+		}
+		d.listenerEnd[listenerName] <- true
 	}
+	d.listenerLock.RUnlock()
+	d.listenerLock.Lock()
+	defer d.listenerLock.Unlock()
```
orozery, line 105: This seems inefficient to me.
I would take a writer lock at `handleClusters`, and `handleListeners`, and remove all locks from `AddListener`, 'AddCluster`, 'RemoveListener`, `RemoveCluster`.
Just add a comment for each function that says the function assumes lock is already held.

tests/k8s.sh
```diff
@@ -80,7 +80,7 @@ function test_k8s {
   kubectl cp $SCRIPT_DIR/../pkg/policyengine/policytypes/examples/allowAll.json gwctl:/tmp/allowAll.json
   kubectl exec -i gwctl -- gwctl create policy --type access --policyFile /tmp/allowAll.json
 
-  # get imported service port
+  # get imported service port
```
orozery, line 83: nit: revert?

orozery (approved):

PR #361, opened 2024-02-29
controlplane/control: Introduce peer manager

orozery:
This PR adds the peerManager which is responsible for monitoring peers, and updating peer status (reachable or not).


pkg/controlplane/control/peer.go
```diff
@@ -0,0 +1,276 @@
...
+	ticker := time.NewTicker(heartbeatInterval)
+	defer ticker.Stop()
+
+	backoffConfig := backoff.NewExponentialBackOff()
+
+	reachable := false
+	reachableCond := metav1.Condition{
+		Type:   v1alpha1.PeerReachable,
+		Status: metav1.ConditionFalse,
+		Reason: "Heartbeat",
```
kfirtoledo, line 86: Where do you update the time?

pkg/controlplane/control/peer.go
```diff
@@ -0,0 +1,276 @@
...
+				backoffConfig.MaxElapsedTime = 0
+			}
+
+			reachable = heartbeatOK
+
+			m.lock.Lock()
+			meta.SetStatusCondition(&m.pr.Status.Conditions, reachableCond)
+			m.lock.Unlock()
+
+			m.statusCallback(m.pr)
```
kfirtoledo, line 115: Can you add a comment why you call the callback

pkg/controlplane/control/peer.go
```diff
@@ -0,0 +1,276 @@
...
+func (m *peerManager) Name() string {
+	return "peerManager"
+}
+
+func (m *peerManager) SetStatusCallback(callback func(*v1alpha1.Peer)) {
+	m.statusCallback = callback
+}
+
+// Start the peer manager.
+func (m *peerManager) Start() error {
```
kfirtoledo, line 163: Where do you call this function?

pkg/controlplane/control/peer.go
orozery: Sure (it's for non-CRD mode, which does not watch Peer status).

pkg/controlplane/control/peer.go
orozery: It's called by the runnable manager, which subscribes here:
https://github.com/clusterlink-net/clusterlink/blob/b41c7721d4260568a40aa1f79fc03e26859f90e8/cmd/cl-controlplane/app/server.go#L205

kfirtoledo (approved):

PR #364, opened 2024-02-29
cl-dataplane: Disable controlplane TLS session keys

orozery:
This PR disables envoy from using TLS session keys when connecting to the controlplane.
Enabling session keys produces big TLS client hello packets, which cause a "buffer full" error on the controlplane's SNI proxy.

### DIFF
--- a/cmd/cl-dataplane/app/envoyconf.go
+++ b/cmd/cl-dataplane/app/envoyconf.go
@@ -92,6 +92,7 @@ static_resources:
       typed_config:
         "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.UpstreamTlsContext
         sni: {{.controlplaneGRPCSNI}}
+        max_session_keys: 0 # TODO: remove once controlplane no longer uses inet.af/tcpproxy
         common_tls_context:
           tls_certificate_sds_secret_configs:
           - name: {{.certificateSecret}}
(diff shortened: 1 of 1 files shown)


elevran (approved):

cmd/cl-dataplane/app/envoyconf.go
```diff
@@ -92,6 +92,7 @@ static_resources:
       typed_config:
         "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.UpstreamTlsContext
         sni: {{.controlplaneGRPCSNI}}
+        max_session_keys: 0
```
kfirtoledo, line 95: It would be better to have TODO remove it after the SNI proxy is removed, or maybe explain why you put it in the code.

praveingk:
Interesting, Thanks for the quick fix. Was this config changed recently which made envoy unstable?

orozery:
> Interesting, Thanks for the quick fix. Was this config changed recently which made envoy unstable?

Just to be clear, this is not a bug in envoy, nor an envoy mis-configuration.

In this case, envoy is the TLS client, and our cl-controlplane is the TLS server.
Our cl-controlplane includes an SNI-based TCP-proxy imported from `inet.af/tcpproxy`.
This SNI proxy works by "peeking" the TLS client hello packet, reading it to a buffer of size 4096 bytes.
The issue was caused because envoy TLS client was sending a legitimate TLS client hello packet of size greater than 4096 bytes.
This caused the "connection reset" error that you have seen, because our SNI proxy was unable to handle this "large" client hello.
This is obviously a bug in the SNI proxy used by our controlplane (and the go-dataplane as well BTW).

The workaround fix of this PR is to configure envoy to disable the use of such large client hello packets, by disabling the "TLS session resume" feature (setting `max_session_keys` to 0).

I don't know why we have not seen this error so far. Perhaps only in certain setups envoy decides to use this defaultly-enabled TLS-sessions feature.

kfirtoledo (approved):

orozery:
Opened up a PR (and issue) to fix the bug in tcpproxy:
https://github.com/inetaf/tcpproxy/pull/41

praveingk (approved):

PR #370, opened 2024-03-03
Release: Introduce GitHub Action for the release and installation script.

kfirtoledo:
This PR includes:
1) GitHub action for the release - the action compiles the binaries files for the release on a different platform (using go cross-compile) and uploads them. In addition, it pushes the project docker images. 
2) Installation script: Downloads the release according to the system configuration, untars the binary files, and moves them to ${HOME}/.local/bin.
3) Update version info- use the git version by adding LDFLAGS to makefile instead of the version.txt file.


.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+      with:
+        repo_token: ${{ secrets.GITHUB_TOKEN }}
+        file: ./hack/clusterlink.sh
+        asset_name: clusterlink.sh
+        tag: ${{ github.ref }}
+        overwrite: true
+    - name: Tag and push ClusterLink images with tag 'latest'
+      run: make push-image
+    - name: Tag and push ClusterLink images latest with tag ${{ github.ref_name }}
+      run: make push-image IMAGE_VERSION=${{ github.ref_name }}
```
zivnevo, line 57: Better fits right after the `make docker-build` step

hack/clusterlink.sh
zivnevo, line 1: Consider renaming to `install-clusterlink.sh` or `clusterlink-install.sh`

hack/clusterlink.sh
```diff
@@ -0,0 +1,74 @@
...
+    ARCH=arm64
+    ;;
+  *)
+    echo "This ${OS_ARCH} architecture isn't supported"
+    exit 1
+    ;;
+esac
+
+filename="clusterlink-${CL_OS}-${CL_ARCH}.tar.gz"
+url="https://github.com/clusterlinl-net/clusterlink/releases/download/${VERSION}/${filename}"
```
zivnevo, line 29:
```suggestion
url="https://github.com/clusterlink-net/clusterlink/releases/download/${VERSION}/${filename}"
```

hack/clusterlink.sh
```diff
@@ -0,0 +1,74 @@
...
+
+# Set version to latest if not define and update the url.
+if [ "${VERSION}" = "" ] ; then
+  VERSION="latest"
+  url="https://github.com/clusterlinl-net/clusterlink/releases/${VERSION}/download/${filename}"
```
zivnevo, line 34:
```suggestion
  url="https://github.com/clusterlink-net/clusterlink/releases/${VERSION}/download/${filename}"
```

hack/clusterlink.sh
```diff
@@ -0,0 +1,74 @@
...
+
+# Open the tar file.
+curl -fsLO ${url}
+tar -xzf "${filename}"
+rm "${filename}"
+
+current_path=$(pwd)/clusterlink
+
+install_path=${HOME}/.local/bin
+mv $current_path/* $install_path
```
zivnevo, line 52: `.local/bin` might not be present in the user's home dir

hack/clusterlink.sh
```diff
@@ -0,0 +1,74 @@
...
+printf "\n"
+printf "\e[1;34m.----.  .----. .-. . .-..-. .-..-.    .----.   .--.  .----.     .---.  .----. .-.   .-..----. .-.   .----..---. .----.\n"
+printf "| {}  \\/  {}  \| |/ \| ||  \`| || |   /  {}  \\ / {} \\ | {}  \   /  ___}/  {}  \|  \`.'  || {}  }| |   | {_ {_   _}| {_  \n"
+printf "|     /\\      /|  .'.  || |\  || \`--.\      //  /\\  \|     /   \\     }\      /| |\ /| || .--' | \`--.| {__  | |  | {__ \n"
+printf "\`----'  \`----' \`-'   \`-'\`-' \`-'\`----' \`----' \`-'  \`-'\`----'     \`---'  \`----' \`-' \` \`-'\`-'    \`----'\`----' \`-'  \`----'\e[0m\n"
+printf "\n\n"
+
+printf "%s has been successfully downloaded.\n" "$filename"
+printf "\n"
+printf "ClusterLink CLI (gwctl and cl-adm) added to your environment path:\n"
```
zivnevo, line 65: I'm not sure I understand what this sentence means. I believe we want to say here that the directory where the CLI is **should** be added to PATH.

zivnevo (changes requested):

hack/clusterlink.sh
kfirtoledo: I meant here that I installed it in the directory (this .local/bin should be in the path), in case not, the next sentence should answer that.

zivnevo (approved):
LGTM

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+    - name: Build and compress linux binaries
+      run: |
+        GOOS=linux GOARCH=amd64 CGO_ENABLED=0 make cli-build BIN_DIR=linux
+        tar -czvf clusterlink-linux-amd64.tar.gz --transform 's/linux/clusterlink/' ./linux
+    - name: Build and compress macOS binaries
+      run: |
+        GOOS=darwin GOARCH=amd64 CGO_ENABLED=0 make cli-build BIN_DIR=os
+        tar -czvf clusterlink-darwin-amd64.tar.gz --transform 's/os/clusterlink/' ./os
```
elevran, line 37: nit: perhaps use `macos` or `darwin` instead of `os` for clarity? also in `osarm64`?
you are using `darwin` in the tar.gz so may want to be consistent

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+      run: make push-image
+    - name: Tag and push ClusterLink images latest with tag ${{ github.ref_name }}
+      run: make push-image IMAGE_VERSION=${{ github.ref_name }}
+    - name: Build and compress linux binaries
+      run: |
+        GOOS=linux GOARCH=amd64 CGO_ENABLED=0 make cli-build BIN_DIR=linux
+        tar -czvf clusterlink-linux-amd64.tar.gz --transform 's/linux/clusterlink/' ./linux
+    - name: Build and compress macOS binaries
+      run: |
+        GOOS=darwin GOARCH=amd64 CGO_ENABLED=0 make cli-build BIN_DIR=os
```
elevran, line 36: does GH actions support `foreach`, shell like, syntax?
It may be clearer to iterate on that (or on concat of "GOOS/GOARCH" pairs split internally in the loop

hack/install_clusterlink.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/sh
+set -e
+
+# Detrmine the OS.
+OS=$(uname)
+if [ "${OS}" = "Darwin" ] ; then
+  CL_OS="os"
```
elevran, line 7: ditto on using a more descriptive name than `os`

hack/install_clusterlink.sh
```diff
@@ -0,0 +1,80 @@
...
+fi
+
+printf "\n Downloading %s from %s ...\n" "$filename" "$url"
+
+if ! curl -o /dev/null -sIf "$url"; then
+    printf "\n%s This version of clusterlonk wasn't found\n" "$url"
+    exit 1
+fi
+
+# Open the tar file.
```
elevran, line 44: looking at [Go installation](https://go.dev/doc/install) (for example) they leave the steps of deciding where to install to (local or system path) to the user with a ~recommendation~ default to system.
Maybe we can follow the same and stop the script after the download stage (which, in the Go case, is also also in the docs, not in a script...).
Do we really need the script at all vs having everything in the documentation like Go?
[kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/) also doesn't have a script, just download instructions.

hack/install_clusterlink.sh
```diff
@@ -0,0 +1,80 @@
...
+printf "ClusterLink CLI (gwctl and cl-adm) has been installed in the following directory:\n"
+printf "\n"
+printf "\t\e[1;33m%s\n\e[0m" "$install_path"
+printf "\n"
+printf "If the ClusterLink command is not in your path, please add it using the following command:\n"
+printf "\n"
+printf "\t\e[1;33mexport PATH=\"\$PATH:%s\"\n\e[0m" "$install_path"
+printf "\n"
+printf "For more information on how to set up ClusterLink in your Kubernetes cluster, please see: \e[4mhttps://cluster-link.net/docs/getting-started\e[0m\n"
+printf "\n"
```
elevran, line 80: nit: newline missing

pkg/versioninfo/variables.go
```diff
@@ -15,30 +15,12 @@
...
 	"time"
 )
 
-// @TODO
-// The placement of version.txt inside this package is not ideal.
-// Unfortunately Go does not allow ".." in embed paths. A possible workaround is to
-// use "go:generate" comment to copy the file from the root to this location (and adding
-// the local version to .gitignore...).
-
-// An alternative to using a fixed file, would be to use a go:generate comment with
```
elevran, line 29: question (curiosity): Did you run into issues using `go:generate` for this, or was `LD_FLAGS` better/easier?

elevran (commented):
Please rethink if we want an automation script or just install instructions. It's basically two steps - download and untar to preferred location (which we currently force instead of leaving to user choice.

pkg/versioninfo/variables.go
kfirtoledo: I didn't want to do 2 operations of go build and go generate that's why I used LD_FLAGS.

hack/install_clusterlink.sh
kfirtoledo: But you can check on your computer which programs use this method (pip, Jupyter..)
```shell
 ls ~/.local/bin/
``` 
I think that most of the users don't really care about it, and if someone does he can move it manually to any location he wants. The script prints in the end the location of the installation.

elevran: I actually don't have a `~/.local/bin/` directory at all.
The above tools are python? Maybe this is a python specific convention?

elevran (approved):

PR #372, opened 2024-03-04
Support including files and expanding commands

elevran:
work in progress

add shortcodes and sample usage
- `{{% expand summary="some text" open="true|false" %}}`
- override docsy `{{% readfile "path" %}}` shortcode to not render content as HTML by default (*must* be called with `%`!)
- sites.md uses `expand` and includes `concepts/data/peer_crd.md` using `readfile` in the inner section

### DIFF
--- a/website/content/en/docs/concepts/sites.md
+++ b/website/content/en/docs/concepts/sites.md
@@ -16,12 +16,15 @@ A `Site` represents a location, such as a Kubernetes cluster, participating in a
 Once a `Site` has been added to a `Fabric`, it can communicate with any other `Site`
  belonging to it. All configuration relating to service sharing (e.g., the exporting
  and importing of Services, and the setting of fine grained application policies) can be
- done with lowered privileges (e.g., by users, such as application owners).
+ done with lowered privileges (e.g., by users, such as application owners). Remote sites are
+ represented by the `Peer` Custom Resource Definition (CRD). Each `Peer` CRD instance
+ defines a remote cluster and the network endpoints of its ClusterLink gateways.
 
 ## Initializing a new Site
 
 {{< notice warning >}}
-Creating a new Site is a **Fabric** administrator level operation and should be appropriately protected.
+Creating a new Site is a **Fabric administrator** level operation and should be appropriately
+ protected.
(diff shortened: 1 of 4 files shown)


michalmalka (approved):
LGTM

website/content/en/docs/concepts/sites.md
```diff
@@ -57,55 +60,102 @@ You will need the CA certificate (but **not** the CA private key) and the site c
...
+ during connection establishment. The name is used by importers in referencing an export
+ (see [here]({{< ref "services" >}}) for details).
+
+Gateway endpoint would typically be a implemented via a `NodePort` or `LoadBalancer`
+ Kubernetes Service. A `NodePort` Service would typically be used in local deployments
+ (e.g., when running in KIND clusters during development) and a `LoadBalancer` Service
+ would be used in Cloud based deployments. These can be automatically configured and
+ created via the [operator CRD]{{< ref "#deploy-clusterlink-via-the-operator-and-clusterlink-crd" >}}.
+ Not having any gateways is an error and will be reported in the `Peer`'s Status.
+ In addition, the Status could display additional Peer conditions, such a `Reachable`
```
kfirtoledo, line 148: why could? it is going to display this..

website/static/files/peer_crd_sample.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: clusterlink.net/v1alpha1
```
kfirtoledo, line 1: You will need later to update it to include the status.

kfirtoledo (approved):
Looks good, minor comments

website/static/files/peer_crd_sample.yaml
elevran: the intent was to show what the user would need to send from yaml to kubectl, not show a full blown CRD response. Status would not be set by the user so it was not noted

website/content/en/docs/concepts/sites.md
elevran: the conditions are not decided yet so I did not want to commit ;-)

PR #376, opened 2024-03-07
pkg/api: Move bindings inside imports

orozery:
This PR moves bindings to be included in the import spec. The bindings API is removed.


pkg/controlplane/rest/import.go
```diff
@@ -31,14 +31,22 @@ type importHandler struct {
...
 	return &v1alpha1.Import{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      imp.Name,
 			Namespace: namespace,
 		},
 		Spec: v1alpha1.ImportSpec{
-			Port:       imp.ImportSpec.Service.Port,
-			TargetPort: imp.Port,
+			Port:       imp.Port,
+			TargetPort: imp.TargetPort,
```
kfirtoledo, line 48: What happens if the targetPort is already used?

pkg/api/types.go
```diff
@@ -87,34 +87,23 @@ type Import struct {
 	Name string
 	// Spec represents the attributes of the import service.
 	Spec ImportSpec
-	// Status field contains the import service status.
-	Status ImportStatus
```
kfirtoledo, line 91: Why did you remove the status?

pkg/api/types.go
```diff
@@ -87,34 +87,23 @@ type Import struct {
...
 }
 
-// ImportSpec contains all the import service attributes.
-type ImportSpec struct {
```
kfirtoledo, line 95: Shouldn't be part of importspec?

pkg/controlplane/rest/import.go
```diff
@@ -31,14 +31,22 @@ type importHandler struct {
 }
 
 func toK8SImport(imp *store.Import, namespace string) *v1alpha1.Import {
+	sources := make([]v1alpha1.ImportSource, len(imp.Peers))
+	for i, pr := range imp.Peers {
+		sources[i].Peer = pr
+		sources[i].ExportName = imp.Name
```
kfirtoledo, line 37: Do I need always to specify the ExportName and ExportNamespace? what happens if we don't?

tests/e2e/k8s/test_basic.go
```diff
@@ -43,7 +43,7 @@ func (s *TestSuite) TestConnectivityCRD() {
 			Name: httpEchoService.Name,
 			Port: 80,
 		}
-		require.Nil(s.T(), cl[1].CreateImportCRD(importedService, cl[0], httpEchoService.Name))
+		require.Nil(s.T(), cl[1].CreateImport(importedService, cl[0], httpEchoService.Name))
```
kfirtoledo, line 47: You didn't want to combine the tests? with CRD or without?

orozery: I will do it in a separate PR that enhances tests.

pkg/controlplane/rest/import.go
orozery: The `Import` API only includes an array of peers.
Here, we translate this array to `ImportSource`, which requires not just the peer, but also the export name and namespace.
In the old API (which is kept here), the export name must match the import name, and there's no namespaces.

pkg/api/types.go
orozery: Since it is not really used.

kfirtoledo (approved):

PR #381, opened 2024-03-10
cmd/deploy: Support container registry flag in deploy peer command.

kfirtoledo:
Support container registry flag in deploy peer command.

### DIFF
--- a/cmd/cl-adm/config/config.go
+++ b/cmd/cl-adm/config/config.go
@@ -41,6 +41,11 @@ const (
 	DataplaneDirectoryName = "dataplane"
 	// GWCTLDirectoryName is the directory name containing gwctl certificates.
 	GWCTLDirectoryName = "gwctl"
+
+	// GHCR is the path to the GitHub container registry.
+	GHCR = "ghcr.io/clusterlink-net"
+	// DefaultRegistry is the default container registry used.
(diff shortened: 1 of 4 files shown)


cmd/cl-adm/config/config.go
```diff
@@ -41,6 +41,8 @@ const (
 	DataplaneDirectoryName = "dataplane"
 	// GWCTLDirectoryName is the directory name containing gwctl certificates.
 	GWCTLDirectoryName = "gwctl"
+	// GhcrContainerRegistry represents the remote container registry name on GitHub Container Registry.
+	GhcrContainerRegistry = "ghcr.io/clusterlink-net"
```
elevran, line 45: nits:
- Should probably be `GHCR` or better yet `GHContainerRegistry` since the `CR` is redundant with `ContainerRegistry`.
- Also, does it make sense to somehow add `Default` to the parameter names (e.g., `DefaultRegistry`)?
  And then use `config.DefaultRegistry` in command line options.
 
  ```suggestion
	// GHCR is the path to the GitHub container registry
	GHCR =  "ghcr.io/clusterlink-net"
	// DefaultRegistry is the default container registry used
	DefaultRegistry = GHCR
 ```

elevran (approved):

PR #382, opened 2024-03-10
cmd/deploy: Support ingress port flag

kfirtoledo:
Support ingress port flag for the deploy peer command.

### DIFF
--- a/cmd/cl-adm/cmd/deploy/deploy_peer.go
+++ b/cmd/cl-adm/cmd/deploy/deploy_peer.go
@@ -91,9 +91,9 @@ func (o *PeerOptions) AddFlags(fs *pflag.FlagSet) {
 			"If true, it will also deploy the ClusterLink instance CRD, which will create the ClusterLink components.")
 	fs.StringVar(&o.Ingress, "ingress", string(apis.IngressTypeLoadBalancer), "Represents the type of service used"+
 		"to expose the ClusterLink deployment (LoadBalancer/NodePort/none). This option is only valid if --autostart is set.")
-	fs.Uint16Var(&o.IngressPort, "ingress-port", apis.ExternalDefaultPort,
+	fs.Uint16Var(&o.IngressPort, "ingress-port", apis.DefaultExternalPort,
 		"Represents the ingress port. By default it is set to 443 for LoadBalancer"+
-			" and a random port in range (3000 to 32767) for NodePort. This option is only valid if --autostart is set.")
+			" and a random port in range (30000 to 32767) for NodePort. This option is only valid if --autostart is set.")
 }
 
 // RequiredFlags are the names of flags that must be explicitly specified.
@@ -168,7 +168,7 @@ func (o *PeerOptions) Run() error {
 			Namespace:         o.Namespace,
 			IngressType:       o.Ingress,
 		}
-		if o.IngressPort != apis.ExternalDefaultPort {
+		if o.IngressPort != apis.DefaultExternalPort {
 			cfg.IngressPort = o.IngressPort
 		}
 		instance, err := platform.K8SClusterLinkInstanceConfig(cfg, "cl-instance")
@@ -184,10 +184,10 @@ func (o *PeerOptions) Run() error {
 		}
 	} else {
 		if o.Ingress != string(apis.IngressTypeLoadBalancer) {
-			fmt.Println("flag --autostart is not set, ignore --ingres flag")
+			fmt.Println("flag --autostart is not set, ignoring --ingres flag")
 		}
-		if o.IngressPort != apis.ExternalDefaultPort {
-			fmt.Println("flag --autostart is not set, ignore --ingres-port flag")
+		if o.IngressPort != apis.DefaultExternalPort {
+			fmt.Println("flag --autostart is not set, ignoring --ingres-port flag")
 		}
 	}
 
--- a/cmd/cl-adm/cmd/deploy/deploy_peer.go
+++ b/cmd/cl-adm/cmd/deploy/deploy_peer.go
@@ -46,6 +46,8 @@ type PeerOptions struct {
 	StartInstance bool
 	// Ingress, represents the type of service used to expose the ClusterLink deployment.
 	Ingress string
+	// IngressPort, represents the port number of the service used to expose the ClusterLink deployment.
+	IngressPort uint16
 	// ContainerRegistry is the container registry to pull the project images.
 	ContainerRegistry string
 }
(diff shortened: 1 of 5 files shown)


cmd/cl-adm/cmd/deploy/deploy_peer.go
```diff
@@ -84,6 +86,8 @@ func (o *PeerOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.Namespace, "start-namespace", app.SystemNamespace,
 		"Namespace where the ClusterLink components are deployed if --start is set.")
 	fs.StringVar(&o.Ingress, "start-ingress", string(apis.IngressTypeLoadBalancer),
+		"Represents the type of service used to expose the ClusterLink deployment (LoadBalancer/NodePort/none), if --start is set.")
+	fs.Uint16Var(&o.IngressPort, "start-ingress-port", 443, "Represents the ingress port."+
```
elevran, line 90: if there are dependencies between flags (e.g., `start-ingress-port` only valid with `--start`) consider checking for those and reporting expectations failures (e.g., logging that ingress port is ignored when defined without start, logging that using default ingress port when start is defined but no flags...)

cmd/cl-adm/cmd/deploy/deploy_peer.go
```diff
@@ -84,6 +86,8 @@ func (o *PeerOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.Namespace, "start-namespace", app.SystemNamespace,
```
elevran, line 86: I'm not too fond of the use of `start` command line option. I think `autostart` or `deploy` would have been better.
Or even better - make it a separate command altogether?

Also, consider dropping the `start-` prefix from variables. There are two options:
- if start is not specified, log a warning that these are ignored; or
- save them into a configmap to be used later
If you make it a separate command, then neither option is required.

cmd/cl-adm/cmd/deploy/deploy_peer.go
elevran: Also, if using NodePort, there are probably checks you need to make (e.g., there's a specific range assigned in k8s for NodePort), or be ready to fail on creating the service and report back to the user.

elevran (commented):
what does it mean to deploy without a start?
What gets created, what doesn't? For example, is it only deploying the operator and then `start` lets you define the CRD to use?
If so, suggest splitting to deploy-operator and deploy-instance (conceptually, choose better command names obviously...)

cmd/cl-adm/cmd/deploy/deploy_peer.go
kfirtoledo: Done for checking the flags, about the nodeport there is a check before  the nodeport creation (see in pkg/bootstrap/platform/k8s.go in this commit)

cmd/cl-adm/cmd/deploy/deploy_peer.go
```diff
@@ -80,13 +82,18 @@ func NewCmdDeployPeer() *cobra.Command {
...
 	fs.StringVar(&o.ContainerRegistry, "container-registry", config.DefaultRegistry,
 		"The container registry to pull the project images.")
+	fs.BoolVar(&o.StartInstance, "autostart", false,
+		"If false, it will deploy only the ClusteLink operator and ClusterLink K8s secrets."+
+			"If true, it will also deploy the ClusterLink instance CRD, which will create the ClusterLink components.")
+	fs.StringVar(&o.Ingress, "ingress", string(apis.IngressTypeLoadBalancer), "Represents the type of service used"+
+		"to expose the ClusterLink deployment (LoadBalancer/NodePort/none). This option is only valid if --autostart is set.")
+	fs.Uint16Var(&o.IngressPort, "ingress-port", apis.ExternalDefaultPort,
+		"Represents the ingress port. By default it is set to 443 for LoadBalancer"+
+			" and a random port in range (3000 to 32767) for NodePort. This option is only valid if --autostart is set.")
```
elevran, line 96:
```suggestion
			" and a random port in range (30000 to 32767) for NodePort. This option is only valid if --autostart is set.")
```

cmd/cl-adm/cmd/deploy/deploy_peer.go
```diff
@@ -169,6 +180,13 @@ func (o *PeerOptions) Run() error {
 		if err != nil {
 			return err
 		}
+	} else {
+		if o.Ingress != string(apis.IngressTypeLoadBalancer) {
+			fmt.Println("flag --autostart is not set, ignore --ingres flag")
```
elevran, line 185:
```suggestion
			fmt.Println("flag --autostart is not set, ignoring --ingres flag")
```

cmd/cl-adm/cmd/deploy/deploy_peer.go
```diff
@@ -169,6 +180,13 @@ func (o *PeerOptions) Run() error {
...
+		}
+		if o.IngressPort != apis.ExternalDefaultPort {
+			fmt.Println("flag --autostart is not set, ignore --ingres-port flag")
```
elevran, line 188:
```suggestion
			fmt.Println("flag --autostart is not set, ignoring --ingres-port flag")
```

cmd/cl-adm/cmd/deploy/deploy_peer.go
```diff
@@ -80,13 +82,18 @@ func NewCmdDeployPeer() *cobra.Command {
 func (o *PeerOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.Name, "name", "", "Peer name.")
 	fs.StringVar(&o.CertDir, "cert-dir", ".", "The directory where the certificates for the fabric and peer are located.")
-	fs.BoolVar(&o.StartInstance, "start", false, "If true, deploy a ClusterLink instance that will create ClusterLink components")
-	fs.StringVar(&o.Namespace, "start-namespace", app.SystemNamespace,
-		"Namespace where the ClusterLink components are deployed if --start is set.")
-	fs.StringVar(&o.Ingress, "start-ingress", string(apis.IngressTypeLoadBalancer),
-		" By default, it is set to 443 for LoadBalancer and a random port (3000 to 32767) for NodePort, if --start is set.")
+	fs.StringVar(&o.Namespace, "namespace", app.SystemNamespace,
```
elevran, line 85: do we need a flag for the operator namespace? It seems that this is for the CL components so the operator is fixed?
Also, shouldn't this have an associated `--autostart` message?

pkg/apis/clusterlink.net/v1alpha1/instance_types.go
```diff
@@ -50,8 +50,8 @@ const (
 )
 
 const (
-	// ExternalPort represents the default value for the external ingress service of the LoadBalancer type.
-	ExternalPort = 443
+	// ExternalDefaultPort represents the default value for the external ingress service of the LoadBalancer type.
+	ExternalDefaultPort = 443
```
elevran, line 54:
```suggestion
	DefaultLoadBalancerPort = 443
```
Prefix default values with `Default` and indicate this is `LoadBalancer` specific.
Obviously needs to change usages as well.

elevran (approved):
some nits you can accept or not, but consider comment on the `namespace` command line option

cmd/cl-adm/cmd/deploy/deploy_peer.go
kfirtoledo: This is only for automation, if someone wants to create the operator in a different namespace (like Openshift) it will not use this CLI.
The CLI also creates the secrets in the namespace so it is always needed.

PR #383, opened 2024-03-10
controlplane/control: Support imports from all namespaces

orozery:
This PR allows creating imported services not just in the clusterlink-system namespace, but in all namespaces.
It also adds a "service created" condition to the Import Status.


config/crds/clusterlink.net_imports.yaml
```diff
@@ -67,6 +67,79 @@ spec:
             - port
             - sources
             type: object
+          status:
+            description: Status represents the import status.
+            properties:
+              conditions:
+                description: Conditions of the import.
+                items:
+                  description: "Condition contains details for one aspect of the current
```
elevran, line 76: Is this text auto-generated or manually created?
It seems to contain generic "boilerplate" which isn't too helpful to users (e.g., `FooStatus` and the known conditions).

config/crds/clusterlink.net_imports.yaml
```diff
@@ -67,6 +67,79 @@ spec:
...
+                      description: message is a human readable message indicating
+                        details about the transition. This may be an empty string.
+                      maxLength: 32768
+                      type: string
+                    observedGeneration:
+                      description: observedGeneration represents the .metadata.generation
+                        that the condition was set based upon. For instance, if .metadata.generation
+                        is currently 12, but the .status.conditions[x].observedGeneration
+                        is 9, the condition is out of date with respect to the current
+                        state of the instance.
```
elevran, line 103:
```suggestion
                        that the condition was set based upon. If .metadata.generation
                        and .status.conditions[x].observedGeneration differ, the condition
                        is out of date with respect to the current state of the instance.
```

config/crds/clusterlink.net_imports.yaml
```diff
@@ -67,6 +67,79 @@ spec:
...
+                      format: int64
+                      minimum: 0
+                      type: integer
+                    reason:
+                      description: reason contains a programmatic identifier indicating
+                        the reason for the condition's last transition. Producers
+                        of specific condition types may define expected values and
+                        meanings for this field, and whether the values are considered
+                        a guaranteed API. The value should be a CamelCase string.
```
elevran, line 112: Also seems auto-generated. Users should not care about the value being CamelCase...

config/crds/clusterlink.net_imports.yaml
```diff
@@ -67,6 +67,79 @@ spec:
...
+                      type: string
+                    status:
+                      description: status of the condition, one of True, False, Unknown.
+                      enum:
+                      - "True"
+                      - "False"
+                      - Unknown
+                      type: string
+                    type:
+                      description: type of condition in CamelCase or in foo.example.com/CamelCase.
```
elevran, line 126: also seems autogenerated with notes to developers, not users.

pkg/apis/clusterlink.net/v1alpha1/import.go
```diff
@@ -51,9 +53,15 @@ type ImportSpec struct {
 	// TODO: add LoadBalancingSpec.
 }
 
+const (
+	// ImportServiceCreated is a condition type for indicating whether the import service was successfully created.
+	ImportServiceCreated string = "ImportServiceCreated"
```
elevran, line 58: nit: since this is part of the ImportStatus, consider shortening to `ServiceCreated`?

pkg/controlplane/control/manager.go
```diff
@@ -15,31 +15,53 @@ package control
...
 import (
 	"context"
+	"crypto/sha256"
+	"errors"
 	"fmt"
 
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
```
elevran, line 24: nit: suggest `k8serrors`, not capitalized. Same as `metav1` not `metaV1`

pkg/controlplane/control/manager.go
```diff
@@ -48,10 +70,62 @@ type Manager struct {
 func (m *Manager) AddImport(ctx context.Context, imp *v1alpha1.Import) error {
 	m.logger.Infof("Adding import '%s/%s'.", imp.Namespace, imp.Name)
 
-	newService := &v1.Service{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      imp.Name,
+	var err error
+	defer func() {
```
elevran, line 74: q: why is the function deferred?

pkg/controlplane/control/manager.go
```diff
@@ -61,87 +135,167 @@ func (m *Manager) AddImport(ctx context.Context, imp *v1alpha1.Import) error {
...
+	return m.addService(ctx, userService)
+}
+
+// DeleteImport removes the listening socket of a previously imported service.
+func (m *Manager) DeleteImport(ctx context.Context, name types.NamespacedName) error {
+	m.logger.Infof("Deleting import '%s/%s'.", name.Namespace, name.Name)
+
+	// delete user service
+	errs := make([]error, 2)
+	errs[0] = m.deleteService(ctx, name)
```
elevran, line 171: consider initializing the `errs` array with a single error (from the `deleteService` in L171) only if it is not a nil error?
Then `append`/`errors.Join()` another error to it in line 176 if that fails.

pkg/controlplane/control/manager.go
```diff
@@ -61,87 +135,167 @@ func (m *Manager) AddImport(ctx context.Context, imp *v1alpha1.Import) error {
...
-	port, err := m.ports.Lease(fullName, imp.Spec.TargetPort)
+	leasedPort, err := m.ports.Lease(name, imp.Spec.TargetPort)
 	if err != nil {
 		return fmt.Errorf("cannot generate listening port: %w", err)
 	}
 
-	if newPort {
-		imp.Spec.TargetPort = port
-		newService.Spec.Ports[0].TargetPort = intstr.FromInt32(int32(port))
+	if imp.Spec.TargetPort == 0 {
```
elevran, line 210: do we support specifying a target port on the import?

elevran (changes requested):

config/crds/clusterlink.net_imports.yaml
orozery: This whole directory is auto-generated.

pkg/apis/clusterlink.net/v1alpha1/import.go
orozery: This const is part of the `v1alpha1` package, so you need to prefix it with `Import`.

pkg/controlplane/control/manager.go
orozery: because it manipulates the returned `err`.

pkg/controlplane/control/manager.go
orozery: This requires me to add `if` clauses. And `errors.Join` already handles `nil`

orozery:
Now including service watching and reconciliation as well.

pkg/apis/clusterlink.net/v1alpha1/import.go
elevran: I meant on the string value, not variable name (i.e., `ImportServiceCreated string = "ServiceCreated"`).

k8s doesn't seem to do that. For example, [Service Types](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) don't have a `Service` prefix (i.e., `ClusterIP` not `ServiceClusterIP`).
Just FYI - don't have a preference here.

pkg/controlplane/control/manager.go
```diff
@@ -43,28 +53,106 @@ func (e exportServiceNotExistError) Error() string {
 		e.name.Namespace, e.name.Name)
 }
 
+type conflictingServiceError struct {
+	service *v1.Service
```
elevran, line 57: you coul just copy the `NamespacedName` and `ManagedBy` label and then let the rest of the Service be gc'ed

pkg/controlplane/control/manager.go
elevran: It could obscure the flow, typically `defer` is used for cleanup, whereas here it is used to manipulate the object.
Just to make the flow explicit, could the same be achieved using an `updateCondistions` func and calling it directly whenever `return err` is also called? It would also help in shortening this function somewhat from its current ~150 lines,

pkg/controlplane/control/port.go
```diff
@@ -35,11 +36,20 @@ const (
 	maxRandomTries = 40
 )
 
+type conflictingTargetPortError struct {
+	port      uint16
+	leaseName types.NamespacedName
+}
+
+func (e conflictingTargetPortError) Error() string {
+	return fmt.Sprintf("port %d is already leased to %v", e.port, e.leaseName)
```
elevran, line 45:
```suggestion
	return fmt.Sprintf("port %d is already in use by service %v", e.port, e.leaseName)
```
I think `leased` might be less clear to users

pkg/controlplane/xds/manager.go
```diff
@@ -129,9 +132,9 @@ func (m *Manager) DeleteExport(name types.NamespacedName) error {
 func (m *Manager) AddImport(imp *v1alpha1.Import) error {
 	m.logger.Infof("Adding import '%s/%s'.", imp.Namespace, imp.Name)
 
-	if imp.Spec.TargetPort == 0 {
+	if m.crdMode && !meta.IsStatusConditionTrue(imp.Status.Conditions, v1alpha1.ImportTargetPortValid) {
```
elevran, line 135: is there a Status field, other than the conditions, which can be tested (e.g., `allocatedPort`) and if not, should we consider adding one?
I think this stems from the fact that we allow TargetPort in the Spec. Can you please re-explain what's the use case for that? Not sure I understand why a user would care for the random port being used in the data plane, they should just care about the Service (virtual) port, which should always be valid since defined on the new VIP.

elevran (approved):
Please see comments, none of which should be a blocker for merging.

pkg/controlplane/control/manager.go
orozery: using defer is the safest, since it always occurs, just like using a `finally` clause.
With your suggestion, a future programmer can fail to call `updateConditions` when a new return is introduced.
Also note that this is an accepted pattern even by the `nonamedreturnlinter` which we have enabled:
https://github.com/firefart/nonamedreturns
`Named errors used in defers are not reported.`

pkg/controlplane/xds/manager.go
elevran: won't you be able to reconstruct it (indirectly) from the Service corresponding to the Import?

PR #388, opened 2024-03-12
controlplane/control: Add export status updates

orozery:
This PR adds an "ExportValid" condition to the Export status.
This condition verifies that exports of local services point to an existing service.
We also add a new authz controller to watch Exports.
It is left for the policy dispatcher to disallow ingress access to invalid exports.


pkg/apis/clusterlink.net/v1alpha1/export.go
```diff
@@ -40,9 +42,15 @@ type ExportSpec struct {
 	Port uint16 `json:"port,omitempty"`
 }
 
+const (
+	// ImportServiceCreated is a condition type for indicating whether the export is valid.
```
kfirtoledo, line 46: Need to fix the comment

pkg/controlplane/control/manager.go
```diff
@@ -138,6 +151,93 @@ func (m *Manager) DeleteImport(ctx context.Context, name types.NamespacedName) e
...
+func (m *Manager) addService(ctx context.Context, service *v1.Service) error {
+	return m.checkExportService(ctx, types.NamespacedName{
+		Namespace: service.Namespace,
+		Name:      service.Name,
+	})
+}
+
+// deleteService deletes a service.
+func (m *Manager) deleteService(ctx context.Context, name types.NamespacedName) error {
+	return m.checkExportService(ctx, name)
```
kfirtoledo, line 225: I Didn't understand why in delete you use checkExportService

pkg/controlplane/control/manager.go
```diff
@@ -164,6 +264,23 @@ func serviceChanged(svc1, svc2 *v1.Service) bool {
 	return false
 }
 
+func conditionChanged(conditions *[]metav1.Condition, cond *metav1.Condition) bool {
```
kfirtoledo, line 267: Maybe this should be in  more common package

kfirtoledo (changes requested):

pkg/controlplane/control/manager.go
orozery: But it used only by controlplane/control

pkg/controlplane/control/manager.go
orozery: Because if a service was deleted, I want to check the relevant export (which will possibly make the export invalid)

kfirtoledo (approved):

PR #416, opened 2024-03-20
cmd/cl-adm: rename "peer" commands to "site"

kfirtoledo:
Rename the "peer" commands to "site" commands in cl-adm


orozery (approved):

elevran:
I think we need to be consistent in using Peer or Site throughout (documentation, CLI commands, and CRDs).
Currently, we are not.
I'm fine with both terms, just make sure it is used all over so not to confuse users.

PR #417, opened 2024-03-20
[website] Move menu items to _index.md front matter

elevran:
... instead of defining the main menu entries in `config.toml`.
Only remaining menu item in the config is GitHub link which does not (and should not) have a corresponding markdown file as it is an external link.


michalmalka (approved):
Looks Good!

PR #421, opened 2024-03-20
v1alpha1/import: Set Round-Robin as the default LBScheme

kfirtoledo:
Set ECMP as the default LBScheme for import

### DIFF
--- a/config/crds/clusterlink.net_imports.yaml
+++ b/config/crds/clusterlink.net_imports.yaml
@@ -36,8 +36,9 @@ spec:
             description: Spec represents the attributes of the imported service.
             properties:
               lbScheme:
-                description: LBScheme is the load-balancing scheme to use (e.g., Random,
-                  Static, ECMP)
+                default: ecmp
+                description: LBScheme is the load-balancing scheme to use (e.g., random,
+                  static, ecmp)
                 type: string
               port:
                 description: Port of the imported service.
--- a/config/crds/clusterlink.net_imports.yaml
+++ b/config/crds/clusterlink.net_imports.yaml
@@ -36,9 +36,9 @@ spec:
             description: Spec represents the attributes of the imported service.
             properties:
               lbScheme:
-                default: ecmp
+                default: round-robin
                 description: LBScheme is the load-balancing scheme to use (e.g., random,
-                  static, ecmp)
+                  static, round-robin)
                 type: string
               port:
                 description: Port of the imported service.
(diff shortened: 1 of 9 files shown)


config/crds/clusterlink.net_imports.yaml
```diff
@@ -36,8 +36,9 @@ spec:
             description: Spec represents the attributes of the imported service.
             properties:
               lbScheme:
```
elevran, line 38: nit: Envoy uses `policy`, not `scheme` for this

config/crds/clusterlink.net_imports.yaml
```diff
@@ -36,8 +36,9 @@ spec:
...
-                description: LBScheme is the load-balancing scheme to use (e.g., Random,
-                  Static, ECMP)
+                default: ecmp
+                description: LBScheme is the load-balancing scheme to use (e.g., random,
```
elevran, line 40: q: if the key is called `lbScheme` in line 38, why refer to `LBScheme` in the description?

config/crds/clusterlink.net_imports.yaml
```diff
@@ -36,8 +36,9 @@ spec:
             description: Spec represents the attributes of the imported service.
             properties:
               lbScheme:
-                description: LBScheme is the load-balancing scheme to use (e.g., Random,
-                  Static, ECMP)
+                default: ecmp
```
elevran, line 39: what is the difference between `ecmp` and `random` modes?
I think `ecmp`  is perhaps too "low-level-network-stuff" for most users and we ought to select a simpler name (e.g., `hash`?)

elevran (commented):
unclear about some of the word choices

config/crds/clusterlink.net_imports.yaml
kfirtoledo: ECMP is implemented as a round-robin I will open an issue for that. @zivnevo

config/crds/clusterlink.net_imports.yaml
kfirtoledo: This is an auto-generated file

config/crds/clusterlink.net_imports.yaml
elevran: May I suggest that we have one PR that sets the default and renames it to round-robin (or `rr`) across all files and samples? I counted 9 affected files.

elevran (approved):
unclear about some of the word choices

zivnevo (approved):
ok

config/crds/clusterlink.net_imports.yaml
kfirtoledo: @zivnevo

zivnevo: Back in the old days, we used Policy to encapsulate all the details the user had to set in a given execution of `gwctl` (client name, service name, scheme and default peer). Now that the scheme is set inside an Import CRD, I don't mind renaming it  as `lbPolicy`. I opened Issue #423 .

PR #426, opened 2024-03-20
controlplane/control: Use shorter system service names

orozery:
This PR changes the system service names to be built using SHA1 (40 hex bytes) instead of SHA256 (64 hex bytes). This is to avoid the DNS name limit of 63 characters.

### DIFF
--- a/pkg/controlplane/control/manager.go
+++ b/pkg/controlplane/control/manager.go
@@ -15,8 +15,9 @@ package control
 
 import (
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
+
+	//nolint:gosec // G505: use of weak cryptographic primitive is fine for service name
+	"crypto/md5"
 	"errors"
 	"fmt"
 	"sync"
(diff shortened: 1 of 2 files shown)


pkg/controlplane/control/manager.go
```diff
@@ -498,7 +501,14 @@ func checkServiceLabels(service *v1.Service, importName types.NamespacedName) er
 }
 
 func systemServiceName(name types.NamespacedName) string {
-	return "import-" + hex.EncodeToString(sha256.New().Sum([]byte(name.Namespace+"/"+name.Name)))
+	//nolint:gosec // G401: use of weak cryptographic primitive is fine for service name
```
kfirtoledo, line 504: Do you still need this lint waiver?

kfirtoledo (approved):

elevran:
only nit is in the commit message: code seems to use md5, not sha1  ;-)

Otherwise looks good to me.
Since you don't require any crypto strength hashing, would generated name with template of `import-<ns>-<import>-` also work? It would require reading back the object after creation to determine the actual name, but maybe sufficient and avoids the use of lint workarounds.

elevran:
> I don't understand this.
> But anyhow, notice that I trim the import name and namespace to 10 characters, so I still need the hash to deal with collisions.

In your case you are assuming that MD5 would not create collisions. That's probabilistically correct over a large enough space. The use of `generatedName` gives the responsibility of avoiding collisions to the API masters which they do based on the actual etcd state at the time of request - no guessing...
My comment was that *if* you need to know the name in advance, then the use of generatedName would require reading the object back to know what name it was given. If you don't care for the actual name, then generatedName is valid. This is what the Pod names (for example) use when you see something like user-given-name-random-characters for Pods in a replicaSet, etc.

All of the above is just FYI, no need to change code

PR #429, opened 2024-03-21
controlplane/authz: use policyengine response fields for connection request

kfirtoledo:
Use policyengine response fields for connection request and add import name to the response


zivnevo (approved):

orozery (commented):
I think you need to edit gwctl import as I'm guessing this breaks the demos.

kfirtoledo:
@orozery PTAL

orozery (approved):

PR #431, opened 2024-03-24
policyengine/PolicyDispatcher: allow to delete unprivileged policy

kfirtoledo:
Add delete unprivileged policy function,
(Delete privileged policy is still missing -can be done by adding finalizer before deleting to the policies or separate access policy CRDS to privileged and non-privileged)

### DIFF
--- a/pkg/controlplane/authz/manager.go
+++ b/pkg/controlplane/authz/manager.go
@@ -26,6 +26,7 @@ import (
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/clusterlink-net/clusterlink/pkg/api"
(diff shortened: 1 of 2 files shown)


pkg/controlplane/authz/manager.go
```diff
@@ -173,7 +173,7 @@ func (m *Manager) AddAccessPolicy(policy *api.Policy) error {
 // DeleteAccessPolicy removes an access policy to allow/deny specific connections.
 // TODO: switch from api.Policy to v1alpha1.Policy.
 func (m *Manager) DeleteAccessPolicy(policy *api.Policy) error {
-	return m.policyDecider.DeleteAccessPolicy(policy)
+	return m.policyDecider.DeleteAccessPolicy(policy.Name)
```
elevran, line 176: Unsure if the name alone is sufficient unless it also encodes the namespace.
Even if it does, we can have the same name in a single namespace (privileged and normal are different CRDs).
Seems safer to pass along the entire object (CRD instance).

Unlike the original policy management code, which only allowed a single entry of a policy name - privileged or normal, with CRDs we can have multiple policies with the same name - across namespaces.

pkg/policyengine/PolicyDispatcher.go
```diff
@@ -43,7 +43,7 @@ type PolicyDecider interface {
 	DeleteLBPolicy(policy *api.Policy) error
 
 	AddAccessPolicy(policy *api.Policy) error
-	DeleteAccessPolicy(policy *api.Policy) error
+	DeleteAccessPolicy(name string) error
```
elevran, line 46: See earlier note regarding use of name alone

pkg/policyengine/PolicyDispatcher.go
```diff
@@ -247,10 +247,7 @@ func (pH *PolicyHandler) AddAccessPolicy(policy *api.Policy) error {
...
 }
 
-func (pH *PolicyHandler) DeleteAccessPolicy(policy *api.Policy) error {
-	connPolicy, err := connPolicyFromBlob(policy.Spec.Blob)
-	if err != nil {
-		return err
-	}
-	return pH.connectivityPDP.DeletePolicy(connPolicy.Name, connPolicy.Privileged)
+func (pH *PolicyHandler) DeleteAccessPolicy(name string) error {
+	nonPrivileged := false
```
elevran, line 251: this seems arbitrary.

elevran (commented):
This seems like an intermediate step towards deleting policies, which still tries to maintain backward compatibility.
Is it really worthwhile, or should we move directly to full CRD instance (allowing duplicate names across namespaces, same name for privileged and normal in the same namespace, etc.)?
Am ok approving if the next iteration of the change is already in the works and you're just breaking up the larger change into smaller changesets.
Also, not the test failure...

elevran (approved):

elevran:
for now, we'll support a single level of policies

PR #446, opened 2024-03-26
Support creating EndpointSlice while importing service

praveingk:


pkg/controlplane/control/manager.go
```diff
@@ -123,6 +192,32 @@ func (m *Manager) AddImport(ctx context.Context, imp *v1alpha1.Import) error {
 	return err
 }
 
+// deleteClDataplane deletes pod to ipToPod list.
```
elevran, line 195: See previous comment regarding watching the EndpointSlice of the cl-dataplane service instead of working at the level of individual Pods.

elevran (commented):
early feedback provided in case you find it useful

pkg/apis/clusterlink.net/v1alpha1/import.go
```diff
@@ -48,6 +48,9 @@ type ImportSpec struct {
 	TargetPort uint16 `json:"targetPort,omitempty"`
 	// Sources to import from.
 	Sources []ImportSource `json:"sources"`
+	// The existing service endpoint to merge the imported service with by
+	// creating an endpointslice with the service name pointed to.
+	Merge string
```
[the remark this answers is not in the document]
elevran: Also, we may want to treat imports with `merge` set differently (assuming it's going to be the rare case, not norm).
- In most cases (no merge) we create a Service with a Selector pointing to our Pods (in privileged mode this needs to be "redirected" in the import namespace using ExternalName reference to a ClusterLink service in `clusterlink-system` since we can't have cross namespace selectors)
- In case of merge, create an additional EndpointSlice.

Or we choose the EndpointSlice strategy for all cases. @orozery - thoughts?

cmd/gwctl/subcommand/import.go
```diff
@@ -31,6 +31,7 @@ type importOptions struct {
 	name  string
 	port  uint16
 	peers []string
+	merge string
```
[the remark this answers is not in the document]
praveingk: Yes, I will be changing to bool

pkg/apis/clusterlink.net/v1alpha1/import.go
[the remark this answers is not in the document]
praveingk: I am changing merge to be a bool and not the import service name. I initially thought it might be needed to keep track of the endpointslice when we perform deletion. But I am realizing we can keep track using a label of the peer.

Upon merge option selected, we would not be creating a service at all right? and only be creating endpointslice.

pkg/controlplane/control/controllers.go
```diff
@@ -17,36 +17,54 @@ import (
...
+			},
+			DeleteHandler: mgr.DeleteImport,
+		})
+		if err != nil {
+			return err
+		}
+	}
+	return controller.AddToManager(controllerManager, &controller.Spec{
+		Name:   "control.endpoint",
+		Object: &v1.Pod{},
```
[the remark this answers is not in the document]
praveingk: Oh yes, Let me try if that works to track a specific endpointslice

pkg/apis/clusterlink.net/v1alpha1/import.go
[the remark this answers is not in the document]
elevran: Correct. Only endpoint slice is needed when merging as the service must already exist

pkg/controlplane/control/manager.go
```diff
@@ -41,12 +46,78 @@ type Manager struct {
...
+			Namespace: imp.Namespace,
+			Labels: map[string]string{
+				"kubernetes.io/service-name":             imp.Spec.Merge,
+				"endpointslice.kubernetes.io/managed-by": cpapp.Name,
+			},
+		},
+		AddressType: discv1.AddressTypeIPv4,
+		Endpoints: []discv1.Endpoint{
+			{
+				Addresses: []string{m.clDataplanePodInfo[imp.Namespace]},
```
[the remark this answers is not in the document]
elevran: Data plane pods should scale horizontally even when deployed per namespace

pkg/controlplane/control/manager.go
```diff
@@ -41,12 +46,78 @@ type Manager struct {
...
+			},
+		},
+	}
+	var oldEndpointslice discv1.EndpointSlice
+	var create bool
+	err := m.client.Get(
```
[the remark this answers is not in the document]
praveingk: Yes you are right. I think I was in two minds of adding multiple imports of same service (different peers) in the same endpointslice, However it would make it complicated to delete. Hence, finally decided to have an endpointslice per import from peer.

pkg/apis/clusterlink.net/v1alpha1/import.go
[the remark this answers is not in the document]
elevran: Need to test if k8s auto deletes all endpoint slices when a service is deleted.

pkg/controlplane/control/manager.go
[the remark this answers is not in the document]
elevran: Not sure I follow why need per peer. It should resolve to clusterlink which takes care of the peer multiplicity

pkg/apis/clusterlink.net/v1alpha1/import.go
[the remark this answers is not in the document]
praveingk: No, endpointslice associated managed by us will not be deleted automatically when a service is deleted.

pkg/controlplane/control/controllers.go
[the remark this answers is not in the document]
praveingk: I tried this, And unable to pass a selector to the controller to track endpointslice with certain labels. Hence, using filter in `addEndpoint` to filter out using `managed-by` label

pkg/controlplane/control/manager.go
```diff
@@ -41,12 +46,78 @@ type Manager struct {
...
 
-// AddImport adds a listening socket for an imported remote service.
-func (m *Manager) AddImport(ctx context.Context, imp *v1alpha1.Import) error {
-	m.logger.Infof("Adding import '%s/%s'.", imp.Namespace, imp.Name)
+func (m *Manager) addImportEndpointSlice(ctx context.Context, imp *v1alpha1.Import) error {
+	m.logger.Infof("Adding import endpointslice '%s/%s'.", imp.Namespace, imp.Name)
+
+	protocol := v1.ProtocolTCP
+	var port32 int32
+	newEndpointslice := discv1.EndpointSlice{
```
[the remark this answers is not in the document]
praveingk: Agree

pkg/controlplane/control/manager.go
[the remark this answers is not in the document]
praveingk: Yes, Correct, I have changed the logic so will not be getting and updating endpointslice, since listener remains the same.

pkg/controlplane/control/manager.go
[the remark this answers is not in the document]
praveingk: Got it, Changed this to `[]Endpoint`

pkg/controlplane/control/manager.go
```diff
@@ -41,12 +46,78 @@ type Manager struct {
...
+	if err != nil {
+		return fmt.Errorf("cannot generate listening port: %w", err)
+	}
+	imp.Spec.TargetPort = port
+	port32 = int32(port)
+	endpointPort := discv1.EndpointPort{
+		Port:     &port32,
+		Protocol: &protocol,
+	}
+	if create {
```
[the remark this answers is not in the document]
praveingk: Removed this logic

pkg/controlplane/control/controllers.go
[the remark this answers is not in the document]
elevran: are you using predicates to filter the events when a label selector can't be added to the controller's watch func?

pkg/apis/clusterlink.net/v1alpha1/import.go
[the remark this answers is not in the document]
praveingk: But here, endpointslice since it's managed by clusterlink, it shouldn't be deleted unless the user deletes the import right. Otherwise, it might be unintended action taken by it.

pkg/controlplane/control/controllers.go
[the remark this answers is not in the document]
praveingk: Yes, I am using `ServiceManagedBy` predicate

praveingk:
@elevran @orozery  I think its ready to be reviewed.

pkg/controlplane/control/manager.go
```diff
@@ -193,10 +233,87 @@ func (m *Manager) AddImport(ctx context.Context, imp *v1alpha1.Import) (err erro
...
+	for i := range endpointSliceList.Items {
+		es := endpointSliceList.Items[i]
+		m.logger.Infof("Updating endpointslice %s", es.Name)
+		es.Endpoints = eslice.Endpoints
+		err := m.client.Update(ctx, &es)
+		if err != nil {
+			m.logger.Errorf("Failed to update endpointslice %s: %v.", appName, err)
+		}
+	}
+}
```
orozery, line 278: I think this entire section should be replaced with the following logic:
1. List all "merge=true" imports.
2. For each such import, make sure the import endpointslices are consistent with the current dataplane endpoints.

Your current code assumes the dataplane endpointslices are fixed (same number of endpoint slices, and names cannot change).
I don't think this assumption is good. Specifically, for example, client can scale up/down the number of dataplane replicas.

I would implement step 1 by either maintaing an in-memory map of all merge-type imports, or maybe even better by moving out the merge field (which is currently in the importspec) to be a label. That way you can use a label selector to list just the merge-type imports.

Step 2 would be simply a call to `addImportEndpointSlice`.

pkg/controlplane/control/manager.go
```diff
@@ -193,10 +233,87 @@ func (m *Manager) AddImport(ctx context.Context, imp *v1alpha1.Import) (err erro
 	return m.addImportService(ctx, imp, userService)
 }
 
+// deleteClDataplane deletes pod to ipToPod list.
+func (m *Manager) deleteEndpoint(esID types.NamespacedName) {
+	if strings.Contains(esID.Name, dpapp.Name) {
```
orozery, line 238: This seems hacky.
I think you should instead check of esID.Name is found in m.endpoints (see my previous comment on making m.endpoints indexed by esname).

pkg/controlplane/control/manager.go
```diff
@@ -193,10 +233,87 @@ func (m *Manager) AddImport(ctx context.Context, imp *v1alpha1.Import) (err erro
 	return m.addImportService(ctx, imp, userService)
 }
 
+// deleteClDataplane deletes pod to ipToPod list.
```
orozery, line 236: fix comment

pkg/controlplane/control/manager.go
```diff
@@ -193,10 +233,87 @@ func (m *Manager) AddImport(ctx context.Context, imp *v1alpha1.Import) (err erro
...
+func (m *Manager) deleteEndpoint(esID types.NamespacedName) {
```
orozery, line 237: I would rename `esID` to either `name`, `es`, `endpoint` or `endpointslice`.

orozery: Also, would rename function to `deleteEndpointSlice` to be more accurate.
Same for `addEndpoint`.

pkg/controlplane/control/manager.go
```diff
@@ -193,10 +233,87 @@ func (m *Manager) AddImport(ctx context.Context, imp *v1alpha1.Import) (err erro
...
+			m.logger.Errorf("Failed to update endpointslice %s: %v.", appName, err)
+		}
+	}
+}
+
 // DeleteImport removes the listening socket of a previously imported service.
 func (m *Manager) DeleteImport(ctx context.Context, name types.NamespacedName) error {
 	m.logger.Infof("Deleting import '%s/%s'.", name.Namespace, name.Name)
 
+	defer m.ports.Release(name)
```
orozery, line 284: I think the entire handling you added should be moved to the end of the function, after the original call to `m.ports.Release`, and hence drop the defer you added here.
If it's a merge-type import, the respective service should not have the label which indicates managed by clusterlink, so the service should not be deleted.

pkg/controlplane/control/manager.go
```diff
@@ -193,10 +233,87 @@ func (m *Manager) AddImport(ctx context.Context, imp *v1alpha1.Import) (err erro
...
+
+	// retrieve endpointslices of the imported service to check if created using merge option
+	labelSelector := labels.Set(map[string]string{
```
orozery, line 287: I think you can drop the wrapping `labels.Set`.

pkg/controlplane/control/manager.go
```diff
@@ -81,7 +87,37 @@ type Manager struct {
...
+
+func (m *Manager) addImportEndpointSlice(ctx context.Context, imp *v1alpha1.Import) error {
+	m.logger.Infof("Adding import endpointslice '%s/%s'.", imp.Namespace, imp.Name)
+
+	protocol := v1.ProtocolTCP
+	port32 := int32(imp.Spec.TargetPort)
+
+	es := discv1.EndpointSlice{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      imp.Name + "-" + uuid.NewString()[:5],
```
orozery, line 104: You should truncate imp.Name so that the entire string will not be more than 63 chars.

pkg/controlplane/control/manager.go
```diff
@@ -81,7 +87,37 @@ type Manager struct {
...
+		},
+		AddressType: discv1.AddressTypeIPv4,
+		Endpoints:   m.endpoints[imp.Namespace],
+	}
+	endpointPort := discv1.EndpointPort{
+		Port:     &port32,
+		Protocol: &protocol,
+	}
+	es.Ports = make([]discv1.EndpointPort, 1)
+	es.Ports[0] = endpointPort
```
orozery, line 119: Can you move this to be part of the `es := ` assignment above?

orozery (changes requested):
Also, need to add some e2e testing (including edge-cases).
I have another PR waiting which you can base upon here: #458 (see `test_import.go`).

pkg/controlplane/control/manager.go
```diff
@@ -81,7 +87,37 @@ type Manager struct {
 	lock            sync.Mutex
 	serviceToImport map[string]types.NamespacedName
 
-	logger *logrus.Entry
+	podLock sync.RWMutex
+	// endpoints stores the IPv4/v6 address of cl-dataplane endpoints per namespace
```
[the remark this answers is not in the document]
praveingk: I see, That's good then. I will remove the namespace from the map. Do we need the esname, For multiple cl-dataplane pods, we might have a single cl-dataplane endpointslice, or multiple?  Since I am going to track only cl-dataplane endpointslice, and its limited to a single namespace, maybe I should just use an endpoint list

pkg/controlplane/control/manager.go
```diff
@@ -81,7 +87,37 @@ type Manager struct {
 	lock            sync.Mutex
 	serviceToImport map[string]types.NamespacedName
 
-	logger *logrus.Entry
+	podLock sync.RWMutex
```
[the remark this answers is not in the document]
praveingk: Will be removing this lock, since will be removing the map

pkg/controlplane/control/manager.go
praveingk: That's the same logic which is implemented by the section.
1) First we retrieve the endpoint slices using `managed-by` label , which is set when we do `merge=true`
2) The subsequent for loop iterates of the `endpointsliceList` to make sure its consistent.

I don't understand the assumptions. What do you mean by dataplane endpoint slices are fixed? We are only tracking cl-dataplane's endpointslice. We do not track the name of the endpointslice, its tracked using `service-name=cl-dataplane` label , which is the right way.

pkg/controlplane/control/manager.go
praveingk: I plan to remove this, since it doesn't do anything

pkg/controlplane/control/manager.go
praveingk: By doing that, won't we be returning an unnecessary `conflicingServiceError` from `deleteImportService` by doing so? I think I will anyway move the section to `deleteImportEndpointSlice`

pkg/controlplane/control/manager.go
praveingk: Can you point me to the reference where its done for other imports, so that its consistent

pkg/controlplane/control/manager.go
praveingk: Sure.

pkg/controlplane/control/manager.go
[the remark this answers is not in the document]
orozery: Why do assume a single endpointslice?
It's up to kubernetes to decide this.
In particular:
```
By default, the control plane creates and manages EndpointSlices to have no more than 100 endpoints each. You can configure this with the --max-endpoints-per-slice [kube-controller-manager](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/) flag, up to a maximum of 1000.
```

pkg/controlplane/control/manager.go
orozery: The `conflictingServiceError` in the case of `DeleteImport` is actually a warning, as it is wrapped by `reconcile.TerminalError`, which means "don't retry".
So you will see an error in the log, and that's it.
There's no definite way to tell if this warning is justified, since the import object is already deleted, so you cannot tell if it was a merge-type import.

pkg/controlplane/control/manager.go
orozery: My bad, I see that endpoint slice names can be up to 253 chars.
Import names are actually limited to 63 (we should add a kubebuilder annotation for it).

pkg/controlplane/control/manager.go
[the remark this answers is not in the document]
praveingk: Are we expecting more than  100-1000 cl-dataplane endpoints per namespace? I feel it seems to complicate the logic, but we can support that requirement. However, would also like to understand what is the usual configuration. 
@elevran What do your thoughts?

pkg/controlplane/control/manager.go
praveingk: Fixed it

pkg/controlplane/control/manager.go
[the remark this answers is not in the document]
praveingk: Done, Maintaining a map with esname and combining it when updating endpoint slices.

praveingk:
> Also, need to add some e2e testing (including edge-cases). I have another PR waiting which you can base upon here: #458 (see `test_import.go`).

Let me add the e2e tests.

orozery:
@praveingk I still see issues in `addImportEndpointSlice` / `updateEndpointSlices`:
1. `addImportEndpointSlice` will unnecessarily create multiple eps per import when re-called for whatever reason.
2. There is a race between `addImportEndpointSlice` and `updateEndpointSlices` that can lead to an import missing out endpoints

I think the following changes are needed:
1. Make `addImportEndpointSlice` idempotent by issuing the following API requests: List import ESP, if not exist create, if exist update if necessary.
2. Change `updateEndpointSlices` to list all merge-type imports, and simply call `addImportEndpointSlice` on each.

As said in previous review, I think that (2) above is easiest if you use a label to denote merge-type imports (instead of adding a `merge bool` to the spec).

praveingk:
@orozery Valid Points!
Based on our discussion will do the following: 

1) Maintain a map[string] `mergeImports` for all merge-type imports
2) updateEndpointSlice will iterate over the `mergeImports` and issue `addImportEndpointSlice` for the import.
3) `addImportEndpointSlice` will list the endpoint slices with name `imp.Name`+`-clusterlink.net` , if its found will update it, or create it afresh.

pkg/apis/clusterlink.net/v1alpha1/import.go
```diff
@@ -51,6 +51,9 @@ type ImportSpec struct {
 	TargetPort uint16 `json:"targetPort,omitempty"`
 	// Sources to import from.
 	Sources []ImportSource `json:"sources"`
+	// Merge indicates if the imported service needs to be merged with an existing
+	// service endpoint by just creating an endpointslice pointing to the dataplane listener.
+	Merge bool `json:"merge"`
```
orozery, line 56: I've been rethinking about this.
In CRD-mode, putting this field as a label will yield a cleaner implementation.
I prefer that we move this from the spec to a label, and then make small changes to:
`controlplane/store/types.go`
`controlplane/rest/import.go`
To persist the labels to the store.
(Also a small change to gwctl to plant a label).

pkg/controlplane/control/manager.go
```diff
@@ -469,6 +592,37 @@ func (m *Manager) deleteImportService(ctx context.Context, service, imp types.Na
...
+	m.endpointLock.Lock()
+	m.dpEndpoints[eslice.Name] = eslice.Endpoints
+	m.endpointLock.Unlock()
+
+	return m.updateEndpointSlices(ctx)
+}
+
+// deleteEndpoint removes endpoint from dpEndpoints and updates the endpointslices.
+func (m *Manager) deleteEndpoint(ctx context.Context, eslice types.NamespacedName) error {
+	if _, ok := m.dpEndpoints[eslice.Name]; !ok {
```
orozery, line 613: Better take the lock before this line.

pkg/controlplane/control/manager.go
```diff
@@ -81,9 +84,114 @@ type Manager struct {
...
+	for _, es := range m.dpEndpoints {
+		endpoints = append(endpoints, es...)
+	}
+	m.endpointLock.Unlock()
+
+	return endpoints
+}
+
+func (m *Manager) updateEndpointSlices(ctx context.Context) error {
+	m.importsLock.Lock()
```
orozery, line 110: copy aside imports and release lock

pkg/controlplane/control/manager.go
```diff
@@ -81,9 +84,114 @@ type Manager struct {
...
+	return m.client.Update(ctx, &oldEs)
+}
+
+func (m *Manager) deleteImportEndpointSlice(ctx context.Context, imp types.NamespacedName) error {
+	// retrieve endpointslices of the imported service to check if created using merge option
+	esName := imp.Name + "-" + AppName
+
+	if _, ok := m.mergeImports[imp.Name]; !ok {
+		return nil
+	}
```
orozery, line 178: I don't think this check is needed.

pkg/controlplane/control/manager.go
```diff
@@ -81,9 +84,114 @@ type Manager struct {
...
+	// update endpoint address and ports
+	oldEs.Endpoints = newEs.Endpoints
+	oldEs.Ports = newEs.Ports
+
+	return m.client.Update(ctx, &oldEs)
+}
+
+func (m *Manager) deleteImportEndpointSlice(ctx context.Context, imp types.NamespacedName) error {
+	// retrieve endpointslices of the imported service to check if created using merge option
+	esName := imp.Name + "-" + AppName
```
orozery, line 174: Can you put this in a function used by both current callers?

pkg/controlplane/control/manager.go
```diff
@@ -81,9 +84,114 @@ type Manager struct {
...
+
+	m.logger.Infof("Deleting endpointslice %s", esName)
+	err := m.client.Delete(ctx, &discv1.EndpointSlice{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      esName,
+			Namespace: imp.Namespace,
+		},
+	})
+	if err != nil {
+		m.logger.Errorf("Failed to delete endpointslice: %v.", err)
```
orozery, line 188: Check if not found, and return nil if so.

orozery (changes requested):
And also, don't forget e2e test (that tests all flows, create/delete import, create/delete dataplane pod/service).

orozery:
@praveingk I think it's ready for review

pkg/controlplane/control/manager.go
```diff
@@ -469,46 +614,235 @@ func (m *Manager) deleteImportService(ctx context.Context, service, imp types.Na
...
+
+func (m *Manager) addImportEndpointSlices(ctx context.Context, imp *v1alpha1.Import) error {
+	// check that import service exists
+	importName := types.NamespacedName{
+		Namespace: imp.Namespace,
+		Name:      imp.Name,
+	}
+
+	if err := m.client.Get(ctx, importName, &v1.Service{}); err != nil {
+		if k8serrors.IsNotFound(err) {
```
praveingk, line 748: The specific scenario we were targeting endpointslice involves, the service could to be created externally by another application asynchronously. Hence, it might be a strong assumption to assume the service should already be present esp in Non-CRD mode.

pkg/controlplane/control/manager.go
```diff
@@ -66,6 +95,39 @@ func (e conflictingServiceError) Error() string {
...
+
+type importEndpointSliceName struct {
+	importName                 string
+	dataplaneEndpointSliceName string
+}
+
+func (n *importEndpointSliceName) Get() string {
+	return fmt.Sprintf(
+		"clusterlink-%d-%s-%s",
+		len(n.importName), n.importName, n.dataplaneEndpointSliceName)
```
praveingk, line 111: Why do we need the length to be prefixed in the name? isnt importName unique in itself?

praveingk (commented):
Great, Mostly fine. Just some comments on the non-crd mode operation, which would fail otherwise

pkg/controlplane/control/manager.go
orozery: This is just to mark this is a "clusterlink" endpoint slice, further removing the already small chance of a collision with a non-clusterlink endpoint slice.

praveingk (commented):
Looks good to me @orozery

orozery:
@praveingk did you test it to see it supports your use-case?

praveingk:
Yes, I have tested it.
________________________________
From: Or Ozeri ***@***.***>
Sent: Monday, April 15, 2024 8:47:37 PM
To: clusterlink-net/clusterlink ***@***.***>
Cc: Pravein Govindan Kannan ***@***.***>; Mention ***@***.***>
Subject: [EXTERNAL] Re: [clusterlink-net/clusterlink] Support creating EndpointSlice while importing service (PR #446)

@praveingk did you test it to see it supports your use-case? — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving this because you were mentioned. Message ID: <clusterlink-net/clusterlink/pull/446/c2058176678@ github. com>
ZjQcmQRYFpfptBannerStart
This Message Is From an External Sender
This message came from outside your organization.
<https://us-phishalarm-ewt.proofpoint.com/EWT/v1/AdhS1Rd-!9lFRGd6aOoXTvn5SRYcrW9TGkv9EfeMUDn8nta1an4Nb1SkIBHu2386VVVQxgTLNIYRkUGGZGQ50I4r3r8XF--PTsaS8EJM5s-1qnnjzJz4tNOQyG5uq2gxYjLg$>
Report Suspicious

ZjQcmQRYFpfptBannerEnd

@praveingk<https://github.com/praveingk> did you test it to see it supports your use-case?

orozery (approved):

PR #453, opened 2024-03-27
Namespace support and PrivilegedAccessPolicy as a separate CRD

zivnevo:
This PR handles the following:
* Separate CRDs for privileged and non-privileged access policies. Privileged policies are cluster-wide, non-privileged policies are namespaced
* PDP only considers non-privileged access policies which are in the same namespace as the Import resource.

**Note**: We consider policies in the same namespace as the **Import** resource, which may live in a different namespace from the **client** .

Left to be done:
* PolicyDispatcher currently only allows adding and deleting regular policies. User has no way to add privileged access policies, not from CRDs, nor from CRUD. This will be addressed in a follow-up PR (probably after the release)
* More testing for multiple-namespace scenarios


pkg/policyengine/PolicyDispatcher.go
```diff
@@ -101,7 +101,8 @@ func (pH *PolicyHandler) decideIncomingConnection(
 	req *connectivitypdp.ConnectionRequest,
 ) (connectivitypdp.ConnectionResponse, error) {
 	dest := getServiceAttrs(req.DstSvcName, "")
-	decisions, err := pH.connectivityPDP.Decide(req.SrcWorkloadAttrs, []connectivitypdp.WorkloadAttrs{dest})
+	decisions, err := pH.connectivityPDP.Decide(req.SrcWorkloadAttrs, []connectivitypdp.WorkloadAttrs{dest},
```
elevran, line 104: is there a similar decideOutgoingConnection using source namespace?

pkg/policyengine/connectivitypdp/connectivity_pdp.go
```diff
@@ -35,22 +36,23 @@ const (
...
 type PDP struct {
 	privilegedPolicies policyTier
 	regularPolicies    policyTier
 }
 
-// policyTier holds a set of ConnectivityPolicies, split into deny policies and allow policies
+// policyTier holds a set of AccessPolicies, split into deny policies and allow policies
 // Within a tier, no two policies can have the same name, even if one is deny and the other is allow.
 type policyTier struct {
+	privileged    bool
```
elevran, line 49: why needed? this is because you want to evaluate one tier before the other, correct?
Doesn't PDP already have two tiers named by their priorities?

pkg/policyengine/connectivitypdp/connectivity_pdp.go
```diff
@@ -35,22 +36,23 @@ const (
...
 // whether a given connection is allowed or denied by these policies.
 type PDP struct {
 	privilegedPolicies policyTier
 	regularPolicies    policyTier
 }
 
-// policyTier holds a set of ConnectivityPolicies, split into deny policies and allow policies
+// policyTier holds a set of AccessPolicies, split into deny policies and allow policies
 // Within a tier, no two policies can have the same name, even if one is deny and the other is allow.
 type policyTier struct {
```
elevran, line 48: won't you need a different structure to hold the Privileged and Regular policies since they're now different structures?

pkg/policyengine/connectivitypdp/connectivity_pdp_test.go
```diff
@@ -42,30 +47,27 @@ func TestPrivilegedVsRegular(t *testing.T) {
 	workloadSet := []v1alpha1.WorkloadSetOrSelector{trivialWorkloadSet}
 	trivialConnPol := v1alpha1.AccessPolicy{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "reg",
+			Name:      "reg",
+			Namespace: defaultNs,
```
elevran, line 51:
```suggestion
			Namespace: defaultNS,
```

pkg/policyengine/connectivitypdp/connectivity_pdp_test.go
```diff
@@ -241,9 +254,20 @@ func addPoliciesFromFile(pdp *connectivitypdp.PDP, filename string) error {
 			return err
 		}
 
-		err = pdp.AddOrUpdatePolicy(&policy)
-		if err != nil {
-			fmt.Printf("invalid connectivity policy: %v\n", err)
+		if policy.Kind == "PrivilegedAccessPolicy" {
```
elevran, line 257: Is the use of the intermediate representation necessary?
I'm not fully clear how much is point in time and how much is final code, post split to different kinds? 😕

pkg/policyengine/examples/allowToReviews.json
```diff
@@ -18,4 +17,4 @@
             }
         ]
     }
-}
+}
```
elevran, line 20: nit: missing new line

pkg/policyengine/examples/allowAll.json
```diff
@@ -14,4 +13,4 @@
             }
         ]
     }
-}
+}
```
elevran, line 16: nit: missing newline

elevran (commented):
early feedback, need to more thoroughly review logic and policy tiers

pkg/apis/clusterlink.net/v1alpha1/accesspolicy.go
```diff
@@ -56,8 +72,6 @@ type WorkloadSetOrSelector struct {
 
 // AccessPolicySpec contains all attributes of an access policy.
```
kfirtoledo, line 73: all attributes of an unprivileged access policy?

kfirtoledo (commented):
Looks good, small comment.

pkg/policyengine/connectivitypdp/connectivity_pdp.go
zivnevo: It's needed because the two tiers work slightly different when it comes to namespaces: the privileged tier doesn't care about policy namespace, while the unprivileged tier does.

pkg/policyengine/connectivitypdp/connectivity_pdp.go
zivnevo: That's why I only hold policy specs, which are the same struct for AccessPolicy and for PrivilegedAccessPolicy

pkg/policyengine/connectivitypdp/connectivity_pdp_test.go
zivnevo: Testing will be improved (as the PR description says). I should probably avoid file-based tests and switch to struct-based

pkg/apis/clusterlink.net/v1alpha1/accesspolicy.go
zivnevo: of both privileged and unprivileged.

pkg/policyengine/examples/allowAll.json
zivnevo: Fixed. Apparently VSCode has a setting for this: `files.insertFinalNewline`. Default is Off, which means it will remove final new lines whenever a file gets saved.

elevran (approved):
thanks for getting this underway - let's update the website docs once the next PR lands, adding support for Privileged in policy evaluation.

Issue #450, opened 2024-03-27
Peer healthcheck strategy


orozery:
Review the current healthcheck strategy.
Do we want to allow per-peer configuration?

elevran:
@orozery - work still needed or has it been refined since #454?

orozery:
I think the current strategy is good enough.
If any other use-case will teach us different, we can re-open or create a new issue.

PR #454, opened 2024-03-27
controlplane/control: Change heartbeat strategy

orozery:
This PR changes the heartbeat timing strategy, to be more similar to the envoy strategy using
a single request per interval, and threshold for changing state.


elevran (approved):
Please consider the default healthyInterval: 1sec might lead to excessive traffic?

orozery:
> Please consider the default healthyInterval: 1sec might lead to excessive traffic?

The time it takes to detect a failed peer: healthyInterval * unhealthyThreshold = 1s * 5 = 5s
The time it takes to detect a peer has resurrected: unhealthyInterval + healthyInterval * (healthyThreshold - 1) = 10s + 1s * 2 = 12s

If we increase `healthyInterval` to, say, 10s, those time increase to 50s, 30s.
I think around 10s is more reasonable for those times, and this is why I prefer to keep it in 1s.

Doing a small TCP round-trip every 1s (for every peer) seems reasonable to me.

If we do decide to increase it, I was thinking we can add a per-peer configuration for these values (with a default).
This will also allow the tests to set a small interval time to avoid time-wasting.

orozery:
@elevran I'm merging as I want to open a defendant PR.
Please see the issue I created here: #450.
We can discuss and later change these values.

Issue #397, opened 2024-03-13
Complete Service concept in documentation


elevran:
Based on new CRD management model.
This includes Import and Export (the page and headings already exist - content is missing)

PR #469, opened 2024-03-31
[website] Service sharing concept documentation

elevran:
Fixes: #397

### DIFF
--- a/website/content/en/docs/concepts/services.md
+++ b/website/content/en/docs/concepts/services.md
@@ -1,7 +1,7 @@
 ---
 title: Services
 description: Sharing Services
-weight: 36
+weight: 30
 ---
 
 <!--
--- a/website/content/en/docs/concepts/services.md
+++ b/website/content/en/docs/concepts/services.md
@@ -4,11 +4,6 @@ description: Sharing Services
 weight: 30
 ---
 
-{{< notice info >}}
-Services sharing is done on a per namespace basis and does not require cluster wide privileges.
- It is intended to be used by application owners having access to their own namespaces only.
-{{< /notice >}}
-
 ClusterLink uses services as the unit of sharing between peers.
  One or more peers can expose an (internal) Kubernetes Service to
  be consumed by other [peers]({{% ref "peers" %}}) in the [fabric]({{% ref "fabric" %}}).
@@ -19,6 +14,11 @@ ClusterLink uses services as the unit of sharing between peers.
  shared by peers in the fabric. Note that the exporting cluster must be
  [configured as a peer]({{% ref "peers#add-or-remove-peers" %}}) of the importing
  cluster.
+
+{{< notice info >}}
+Services sharing is done on a per namespace basis and does not require cluster wide privileges.
+ It is intended to be used by application owners having access to their own namespaces only.
+{{< /notice >}}
 
 A service is shared using a logical name. The logical name does not have to match
  the actual Kubernetes Service name in the exporting cluster. Exporting a service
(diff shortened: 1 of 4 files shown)


website/content/en/docs/concepts/services.md
```diff
@@ -1,26 +1,204 @@
...
+  - *ExportName* (string, required): name of the remote export.
+- **LBScheme** (string, optional): load balancing method to select between different
+ Sources defined. The default policy is `random`, but you could override it to use
+ `round-robin` or `static` (fixed) assignment.
+
+<!-- Importing multiport? It is not possible... Could use merge in future?
+ perhaps, but might requires explicit service name so can merge correctly
+ or use port set instead of individual port per export/import -->
+
+As with exports, importing a Service does not automatically make is accessible by
```
michalmalka, line 175: is should be it

michalmalka (approved):

website/content/en/docs/concepts/services.md
```diff
@@ -1,26 +1,204 @@
 ---
 title: Services
 description: Sharing Services
-weight: 36
+weight: 30
 ---
 
-<!--
-Each task should give the user
+{{< notice info >}}
```
kfirtoledo, line 7: Maybe this info message should be after the text and you introduce service

website/content/en/docs/concepts/services.md
```diff
@@ -1,26 +1,204 @@
...
+}
+```
+
+{{% /expand %}}
+
+The ExportSpec defines the following fields:
+
+- **Host** (string, optional): the name of the service being exported. If empty,
+ the export shall refer to a Kubernetes Service with the same name as the instance's
+ `metadata.name`. Currently, only TCP based services can be exported[^UDP]. It is an
```
kfirtoledo, line 86: actually is `metadata.name.`metadata.nnamespace.svc.cluster.local

website/content/en/docs/concepts/services.md
```diff
@@ -1,26 +1,204 @@
...
+type ExportStatus struct {
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+```
+
+{{% /expand %}}
+
+The ExportSpec defines the following fields:
+
+- **Host** (string, optional): the name of the service being exported. If empty,
```
kfirtoledo, line 84: We can support DNS and IPs

website/content/en/docs/concepts/services.md
```diff
@@ -1,26 +1,204 @@
...
+```
+
+{{% /expand %}}
+
+The ImportSpec defines the following fields:
+
+- **Port** (integer, required): the imported, user facing, port number define
+ on the created service object.
+- **TargetPort** (integer, optional): this is the internal listening port
+ used by the data plane pods to represent the remote services. Typically the
```
kfirtoledo, line 154:
```suggestion
 used by the ClusterLink data plane pods to represent the remote services.  Typically the
```

website/content/en/docs/concepts/services.md
```diff
@@ -1,26 +1,204 @@
...
+ as is done for NodePort services.
+- **Sources** (source array, required): references to remote exports providing backends
+ for the Import. Each reference names a different export through the combination of
+  - *Peer* (string, required): name of ClusterLink peer where the export is defined.
+  - *ExportNamespace* (string, required): name of the namespace on the remote peer where
+   the export is defined.
+  - *ExportName* (string, required): name of the remote export.
+- **LBScheme** (string, optional): load balancing method to select between different
+ Sources defined. The default policy is `random`, but you could override it to use
+ `round-robin` or `static` (fixed) assignment.
```
kfirtoledo, line 169: The default is round-robin like in K8s as @zivnevo suggested

website/content/en/docs/concepts/services.md
```diff
@@ -1,26 +1,204 @@
...
+
+<!-- Importing multiport? It is not possible... Could use merge in future?
+ perhaps, but might requires explicit service name so can merge correctly
+ or use port set instead of individual port per export/import -->
+
+As with exports, importing a Service does not automatically make it accessible by
```
kfirtoledo, line 175: Maybe need to add that the connection is not open yet, just when traffic arrives.

kfirtoledo (commented):
Looks good, some small comments

website/content/en/docs/concepts/services.md
elevran: this refers to the name inside the same namespace.
I think it should be clear since Service is always namespaced.

website/content/en/docs/concepts/services.md
elevran: true, but we should try to keep users on the "simple path" of using services and not introduce more paths for testing just yet.

website/content/en/docs/concepts/services.md
elevran: it should be clear since we write about service and needing to add a policy to it

kfirtoledo (approved):

PR #489, opened 2024-04-03
Create config.yml to control issue templates

elevran:


michalmalka (commented):
Is this going to disable the option of non bug/feature issue? 
what's about other issue types - like documentation 
and probably there are more...

elevran:
> Is this going to disable the option of non bug/feature issue? what's about other issue types - like documentation and probably there are more...

Yes, it will. We can add an empty template of have specific templates for whatever we want (question, documentation, etc. we cam have a "something else" template that's blank - template content it's up to us - and have it show up as another button instead of a link at the bottom).
Or we can NOT merge this and keep the option to create blank issues as fallback
Your call to approve or close

PR #490, opened 2024-04-03
Unified handling of AccessPolicies and PrivilegedAccessPolicies.

zivnevo:
Introducing an abstract interface to allow handling AccessPolicies and PrivilgedAccessPolicies in a unified way.


orozery:
> The interface, called `SetsAccess`, is defined in the `connectionpdp` package rather than in `v1alpha1`, as kubebuilder doesn't seem to like interfaces in the code it scans.

I actually prefer it is defined in connectionpdp :satisfied:

orozery:
@zivnevo I prefer to keep `pkg/apis` as neat as possible.
You moved the interface out of there, but I would also get rid of the interface implementation functions you introduced there.

My suggestion:
Instead of defining the `SetAccess` interface, change the signature of `AddAccessPolicy`
from `AddAccessPolicy(policy connectivitypdp.SetsAccess) error`
to `AddAccessPolicy(policy v1alpha1.AccessPolicy, privileged bool) error`

And change `PrivilegedAccessPolicy` definition to:
```
type PrivilegedAccessPolicy struct {
	AccessPolicy   `json:",inline"`
}
```

elevran:
@zivnevo I think you need `make codegen` and then commit and upload the newly generated files as the description comments have changed

pkg/apis/clusterlink.net/v1alpha1/accesspolicy.go
```diff
@@ -35,20 +34,22 @@ type PrivilegedAccessPolicy struct {
...
+type PrivilegedAccessPolicy struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
 	// Spec represents the attributes of the exported service.
 	Spec AccessPolicySpec `json:"spec,omitempty"`
 }
 
+// Important note: In the code, PrivilegedAccessPolicy is unmarshalled into AccessPolicy.
+// The two structs must be therefore identical, or something may break.
```
elevran, line 51:
```suggestion
// Therefore, the two structs must be identical, or something may break.
```

pkg/controlplane/authz/controllers.go
```diff
@@ -31,10 +31,24 @@ func CreateControllers(mgr *Manager, controllerManager ctrl.Manager, crdMode boo
 			Name:   "authz.access-policy",
 			Object: &v1alpha1.AccessPolicy{},
 			AddHandler: func(ctx context.Context, object any) error {
-				return mgr.addAccessPolicy(object.(*v1alpha1.AccessPolicy))
+				return mgr.AddAccessPolicy(object.(*v1alpha1.AccessPolicy), false)
 			},
 			DeleteHandler: func(ctx context.Context, name types.NamespacedName) error {
-				return mgr.deleteAccessPolicy(name)
+				return mgr.DeleteAccessPolicy(name)
```
elevran, line 37: shouldn't this also have a bool flag to differentiate privileged and normal?
can a user specify a `metadata.namespace` for privileged access policies as well? 
The metadata field allows it, but would k8s API reject it as invalid or ignore the namespace (but retain it when Getting)?

I guess I'm confused by why it's needed in `AddAccessPolicy` but not in other methods.

pkg/controlplane/rest/accesspolicy.go
```diff
@@ -29,6 +32,17 @@ func accessPolicyToAPI(policy *store.AccessPolicy) *api.Policy {
 	return &policy.Policy
 }
 
+// accessPolicyFromBlob unmarshals an AccessPolicy object encoded as json in a byte array.
```
elevran, line 35: I understand that defining Privileged in terms of normal access policy can be useful (e.g., less code), but it also creates additional code (e.g., marshalling blobs instead of via apimachinery directly). It's somewhere in the area of "it's almost the same thing, but only 90% so..." so I wonder if you want to keep them more separate or not in the code (structures and logic paths). I don't have a strong opinion, but think it could be somewhat confusing for someone who sees the code for the first time (e.g., the note about the two structs have to be the same or things break...)

pkg/apis/clusterlink.net/v1alpha1/accesspolicy.go
elevran: I wonder if there is some type assertion / cast that you can do to validate it in code so the compiler complains. I've seen it done for structs that implement an interface and must continue to do so as the interface evolves

pkg/policyengine/PolicyDispatcher.go
```diff
@@ -248,19 +236,10 @@ func (pH *PolicyHandler) DeleteLBPolicy(policy *api.Policy) error {
 	return pH.loadBalancer.DeletePolicy(lbPolicy)
 }
 
-func (pH *PolicyHandler) AddAccessPolicy(policy *api.Policy) error {
-	connPolicy, err := connPolicyFromBlob(policy.Spec.Blob)
-	if err != nil {
-		return err
-	}
-	return pH.connectivityPDP.AddOrUpdatePolicy(connPolicy)
+func (pH *PolicyHandler) AddAccessPolicy(policy *crds.AccessPolicy, privileged bool) error {
```
elevran, line 239: nit: keep variable lower case as you're not using full words anyway? (`ph *PolicyHandler` works for me the same as `pH *PolicyHandler`)

pkg/policyengine/connectivitypdp/connectivity_pdp.go
```diff
@@ -101,38 +101,28 @@ func (pdp *PDP) GetPolicies() []v1alpha1.AccessPolicy {
...
-	pdp.regularPolicies.addPolicy(policyName, &policy.Spec)
 	return nil
 }
 
 // DeletePolicy deletes an AccessPolicy with the given name and privilege from the PDP.
 // If no such AccessPolicy exists in the PDP, an error is returned.
-func (pdp *PDP) DeletePolicy(policyName types.NamespacedName, privileged bool) error {
-	if privileged {
+func (pdp *PDP) DeletePolicy(policyName types.NamespacedName) error {
+	if policyName.Namespace == "" { // privileged policies have no namespace, regular policies do
```
elevran, line 125: Please confirm that k8s indeed removes it since the metadata has a field for it regardless.

elevran (approved):
- have some questions regarding assumption of namespace always being empty for Privileged.
- net-net, unsure if trying to reuse the same format for both is worth it inside the PDP. Is it really worth it?

zivnevo:
> @zivnevo I prefer to keep `pkg/apis` as neat as possible. You moved the interface out of there, but I would also get rid of the interface implementation functions you introduced there.
> 
> My suggestion: Instead of defining the `SetAccess` interface, change the signature of `AddAccessPolicy` from `AddAccessPolicy(policy connectivitypdp.SetsAccess) error` to `AddAccessPolicy(policy v1alpha1.AccessPolicy, privileged bool) error`
> 
> And change `PrivilegedAccessPolicy` definition to:
> 
> ```
> type PrivilegedAccessPolicy struct {
> 	AccessPolicy   `json:",inline"`
> }
> ```

I tried the above suggestion as well as simply having `type PrivilegedAccessPolicy AccessPolicy` but that didn't work well: when I ran `make codegen` KubeBuilder didn't create a CRD for PrivilegedAccessPolicy. Maybe it requires the struct to have the TypeMeta and ObjectMeta fields?

I therefore left the definition of `PrivilegedAccessPolicy` as is, and added a comment that the two structs must remain identical. Throughout the code we only use AccessPolicy with an additional boolean `privileged` flag. The only place where PriviligedAccessPolicy is recognized is in its controller. The controller will unmarshal PriviligedAccessPolicy into AccessPolicy, and will set the `privileged` flag to `true`. I think it's a reasonable compromise between readability and maintainability, but any other suggestions are welcome.

orozery:
@zivnevo new suggestion, which is somewhat an optimization on your first approach:
Define a member function `GetSpec() *AccessPolicySpec` for each of the 2 types (this is similar to `metav1.ObjectMeta` which defines `GetName`, `GetNamespace` etc).

Outside of `pkg/apis`, define this interface:
```
type Object interface {
	metav1.Object
        GetSpec() *v1alpha1.AccessPolicySpec
}
```

I can also live with adding `IsPrivileged` function, but at least we saved adding the `GetName` and `GetNamespace`.

zivnevo:
> @zivnevo new suggestion, which is somewhat an optimization on your first approach: Define a member function `GetSpec() *AccessPolicySpec` for each of the 2 types (this is similar to `metav1.ObjectMeta` which defines `GetName`, `GetNamespace` etc).
> 
> Outside of `pkg/apis`, define this interface:
> 
> ```
> type AccessPolicy interface {
> 	metav1.Object
>         GetSpec() *v1alpha1.AccessPolicySpec
> }
> ```
> 
> I can also live with adding `IsPrivileged` function, but at least we saved adding the `GetName` and `GetNamespace`.

Another option, which I'm currently leaning towards, is to add to the PDP package an internal (non-CRD) AccessPolicy struct with a `privileged` field, just like we had before we started with CRDs. The controllers will translate the two CRDs into this struct, and pass it on to manager, policy-dispatcher, pdp. Then, we need no interface and no extra code in `v1alpha1`

pkg/policyengine/connectivitypdp/connectivity_pdp.go
zivnevo: Added back the `privileged` parameter to this function

pkg/controlplane/authz/controllers.go
zivnevo: I put the flag back in place

pkg/apis/clusterlink.net/v1alpha1/accesspolicy.go
zivnevo: Introduced a new opaque, PDP-internal, generalized representation of AccessPolicy and PrivilegedAccessPolicy CRDs.
We no longer need this assertion

pkg/controlplane/rest/accesspolicy.go
zivnevo: REST API is a throw-away code, so probably no need to make it extra pretty.

zivnevo:
> > Another option, which I'm currently leaning towards, is to add to the PDP package an internal (non-CRD) AccessPolicy struct with a `privileged` field, just like we had before we started with CRDs. The controllers will translate the two CRDs into this struct, and pass it on to manager, policy-dispatcher, pdp. Then, we need no interface and no extra code in `v1alpha1`
> 
> You can re-use `v1alpha1.AccessPolicySpec` as a field in the type defined in the pdp package, instead of duplicating it.

I made the above change as you suggested. PTAL

orozery (approved):

pkg/policyengine/connectivitypdp/accesspolicy.go
```diff
@@ -0,0 +1,44 @@
...
+
+package connectivitypdp
+
+import (
+	"github.com/clusterlink-net/clusterlink/pkg/apis/clusterlink.net/v1alpha1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// AccessPolicy is an opaque, PDP-internal, generalized representation of AccessPolicy and PrivilegedAccessPolicy CRDs.
+type AccessPolicy struct {
```
elevran, line 22: nit: if this is an internal/opaque representation, there should not be a reason to export the type.

pkg/policyengine/connectivitypdp/accesspolicy.go
```diff
@@ -0,0 +1,44 @@
...
+	name       types.NamespacedName
+	privileged bool
+	spec       v1alpha1.AccessPolicySpec
+}
+
+// PolicyFromCRD converts the AccessPolicy CRD into the PDP's AccessPolicy.
+func PolicyFromCRD(vap *v1alpha1.AccessPolicy) *AccessPolicy {
```
elevran, line 29: nit: may want to use `PolicyFromCR` (i.e., without the `definition`) - you're converting from a custom resource and not from a custom resource *definition*

pkg/policyengine/connectivitypdp/connectivity_pdp_test.go
```diff
@@ -73,7 +73,7 @@ func TestPrivilegedVsRegular(t *testing.T) {
 	require.Equal(t, connectivitypdp.DefaultDenyPolicyName, decisions[0].MatchedBy)
 	require.Equal(t, false, decisions[0].PrivilegedMatch)
 
-	err = pdp.AddOrUpdatePolicy(&trivialConnPol, false)
+	err = pdp.AddOrUpdatePolicy(connectivitypdp.PolicyFromCRD(&trivialConnPol))
```
elevran, line 76: would you be able to hide the internal rep if the code accepted `interface{}` (or using generics) and then you internally call the right function (privileged or not) based on type assertion?

elevran (approved):

PR #499, opened 2024-04-07
docs/getting-started: update the uninstall section

kfirtoledo:
Update the uninstall section for ClusterLink CLI

### DIFF
--- a/website/content/en/docs/getting-started/users.md
+++ b/website/content/en/docs/getting-started/users.md
@@ -115,4 +115,12 @@ To deploy ClusterLink on another cluster, please repeat steps 2-3 in a console w
 ## Try it out
 
 Check out the [ClusterLink Tutorials]({{< ref "tutorials" >}}) for setting up
- multi-cluster connectivity for applications using two or more clusters.
+multi-cluster connectivity for applications using two or more clusters.
+
+## Uninstall ClusterLink
+
+To uninstall the ClusterLink CLI, use the following command:
+
+```sh
+rm `which clusterlink`
+```


website/content/en/docs/getting-started/users.md
```diff
@@ -115,4 +115,25 @@ To deploy ClusterLink on another cluster, please repeat steps 2-3 in a console w
...
- multi-cluster connectivity for applications using two or more clusters.
+multi-cluster connectivity for applications using two or more clusters.
+
+## Uninstall ClusterLink
+
+To uninstall the ClusterLink CLI, use the following command:
+
+{{< tabpane text=true >}}
+{{% tab header="Regular User" %}}
+
```
elevran, line 126: both root and non-root can be combined with 
```sh
$ rm `which clusterlink`
```

website/content/en/docs/getting-started/users.md
```diff
@@ -115,4 +115,25 @@ To deploy ClusterLink on another cluster, please repeat steps 2-3 in a console w
 ## Try it out
 
 Check out the [ClusterLink Tutorials]({{< ref "tutorials" >}}) for setting up
- multi-cluster connectivity for applications using two or more clusters.
+multi-cluster connectivity for applications using two or more clusters.
+
+## Uninstall ClusterLink
```
elevran, line 120: should we also include instructions for uninstalling CRDs, namespaces and the operator?

website/content/en/docs/getting-started/users.md
elevran: also, not that the installation steps did not differentiate between `root` and non-`root` users so it's a bit surprising to make the distinction on uninstall, IMHO

elevran (approved):

website/content/en/docs/getting-started/users.md
kfirtoledo: Can be done in different PR, but because it is more complicated maybe it should be done by ClusterLink CLI?

elevran: deploy and undeploy of all should be done in the same way (e.g., manual or by CLI). It probably means new command for the CLI if we decide to take that route. I **think** I've seen mostly instructions for manual installation of CRDs but that's not a statistically significant study, just (biased?) recollection

elevran: I'm fine with it being a separate PR (esp if you prefer the CLI way). We should prioritize the documentation of the manual step first, though.

PR #507, opened 2024-04-09
cmd/clusterlink: Move file creation to deploy command

kfirtoledo:
1) Move file creations to deploy command.
2) Add a dry-run flag.

### DIFF
--- a/cmd/clusterlink/cmd/deploy/deploy_peer.go
+++ b/cmd/clusterlink/cmd/deploy/deploy_peer.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"path/filepath"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -34,6 +35,7 @@ import (
 	"github.com/clusterlink-net/clusterlink/cmd/clusterlink/config"
 	configFiles "github.com/clusterlink-net/clusterlink/config"
 	apis "github.com/clusterlink-net/clusterlink/pkg/apis/clusterlink.net/v1alpha1"
+	"github.com/clusterlink-net/clusterlink/pkg/bootstrap"
 	"github.com/clusterlink-net/clusterlink/pkg/bootstrap/platform"
 )
 
@@ -59,6 +61,17 @@ type PeerOptions struct {
 	ContainerRegistry string
 	// Tag represents the tag of the project images.
 	Tag string
+	// Dataplanes is the number of dataplanes to create.
+	Dataplanes uint16
+	// DataplaneType is the type of dataplane to create (envoy or go-based)
+	DataplaneType string
+	// LogLevel is the log level.
+	LogLevel string
+	// CRDMode indicates whether to run a k8s CRD-based controlplane.
+	// This flag will be removed once the CRD-based controlplane feature is complete and stable.
+	CRDMode bool
+	// DryRun doesn't deploy the ClusterLink operator but creates the ClusterLink instance YAML.
+	DryRun bool
 }
 
 // NewCmdDeployPeer returns a cobra.Command to run the 'deploy peer' subcommand.
@@ -108,6 +121,15 @@ func (o *PeerOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringToStringVar(&o.IngressAnnotations, "ingress-annotations", nil, "Represents the annotations that"+
 		"will be added to ingress services.\nThe flag can be repeated to add several annotations.\n"+
 		"For example: --ingress-annotations <key1>=<value1> --ingress-annotations <key2>=<value2>.")
+
+	fs.Uint16Var(&o.Dataplanes, "dataplanes", 1, "Number of dataplanes.")
+	fs.StringVar(&o.DataplaneType, "dataplane-type", platform.DataplaneTypeEnvoy,
+		"Type of dataplane, Supported values: \"envoy\" (default), \"go\"")
+	fs.StringVar(&o.LogLevel, "log-level", "info",
+		"The log level. One of fatal, error, warn, info, debug.")
+	fs.BoolVar(&o.DryRun, "dry-run", false,
+		"Dry-run doesn't deploy the ClusterLink operator but creates the ClusterLink instance YAML")
+	fs.BoolVar(&o.CRDMode, "crd-mode", false, "Run a CRD-based controlplane.")
 }
 
 // RequiredFlags are the names of flags that must be explicitly specified.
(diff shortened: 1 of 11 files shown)


cmd/clusterlink/cmd/deploy/deploy_peer.go
```diff
@@ -103,6 +116,14 @@ func (o *PeerOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.Uint16Var(&o.IngressPort, "ingress-port", apis.DefaultExternalPort,
 		"Represents the ingress port. By default it is set to 443 for LoadBalancer"+
 			" and a random port in range (30000 to 32767) for NodePort.\nThis option is only valid if --autostart is set.")
+	fs.StringVar(&o.DataplaneType, "dataplane", platform.DataplaneTypeEnvoy,
+		"Type of dataplane, Supported values: \"envoy\" (default), \"go\"")
```
elevran, line 120: might be able to use tick for string delimiter and avoid escaping quotes

cmd/clusterlink/cmd/deploy/deploy_peer.go
```diff
@@ -103,6 +116,14 @@ func (o *PeerOptions) AddFlags(fs *pflag.FlagSet) {
...
+	fs.Uint16Var(&o.DataplaneReplicas, "dataplane-replicas", 1, "Number of dataplanes.")
+	fs.StringVar(&o.LogLevel, "log-level", "info",
+		"The log level. One of fatal, error, warn, info, debug.")
+	fs.BoolVar(&o.DryRun, "dry-run", false,
```
elevran, line 124: this parameter should be in the documentation (replaces `autostart`?)

cmd/clusterlink/cmd/deploy/deploy_peer.go
```diff
@@ -103,6 +116,14 @@ func (o *PeerOptions) AddFlags(fs *pflag.FlagSet) {
...
+		"Dry-run doesn't deploy the ClusterLink operator but creates the ClusterLink instance YAML")
+	fs.BoolVar(&o.CRDMode, "crd-mode", false, "Run a CRD-based controlplane.")
```
elevran, line 126: still needed? when can it be removed?

elevran: Regardless, shouldn't CRD be defaulted to `true`?

elevran (commented):
documentation should be updated to document new / modified behaviors

cmd/clusterlink/cmd/deploy/deploy_peer.go
kfirtoledo: Actually is not really needed for the operator, because it supports only CRD mode. It is for the k8s.yaml and in the next version we will remove it.

cmd/clusterlink/cmd/deploy/deploy_peer.go
kfirtoledo: check in the documentation if it is not clear enough,
without any flag: just the operator start
--autostart deploy also the instance YAML to the operator so the CLusterLink components start.
--dryrun doesn't do anything just create the YAMLs (doesn't deploy the operator).

cmd/clusterlink/cmd/deploy/deploy_peer.go
kfirtoledo: I prefer to use it like that because cobra print also the default value with "

kfirtoledo:
@elevran PTAL

cmd/clusterlink/cmd/deploy/deploy_peer.go
elevran: since dry-run and auto-start are mutually exclusive (ie I should not define both) maybe combine into a single parameter?
`--components`: `operator` (default), `all` (operator and clusterlink), `none` (generate YAMLs only)
Not sure if `components` is the right flag name (maybe `deploy`?), but you get the idea, I hope.

kfirtoledo: Done, please check the commit https://github.com/clusterlink-net/clusterlink/pull/507/commits/0c0f655af429a5d72f1262bb5428ea0de5c50736

cmd/clusterlink/cmd/deploy/deploy_peer.go
```diff
@@ -47,8 +59,9 @@ type PeerOptions struct {
 	Namespace string
 	// CertDir is the directory where the certificates for the fabric and peer are located.
 	CertDir string
-	// StartInstance, if set to true, deploys a ClusterLink instance that will create the ClusterLink components.
-	StartInstance bool
+	// StartInstance, represents which component to deploy:
+	// `all` (clusterlink components and operator), `operator`, or `none`.
+	StartInstance string
```
elevran, line 64: `Components`?
I think `if StartInstance == "StartNone {` might be ne as readable.

website/content/en/docs/concepts/peers.md
```diff
@@ -76,7 +76,7 @@ Before proceeding, ensure that the CA certificate (the CA private key is not nee
 Install the ClusterLink operator by running the following command:
 
 ```sh
-clusterlink deploy peer --autostart --name <peer_name> --fabric <fabric_name>
+clusterlink deploy peer --name <peer_name> --fabric <fabric_name>
```
elevran, line 79: should you add a note that this starts all and the user can also set `--start` explicitly?

website/content/en/docs/getting-started/users.md
```diff
@@ -54,19 +54,19 @@ To set up ClusterLink on a Kubernetes cluster, follow these steps:
...
 
    ```sh
-   clusterlink deploy peer --autostart --name <peer_name> --fabric <fabric_name>
+   clusterlink deploy peer --name <peer_name> --fabric <fabric_name>
    ```
 
    This command will deploy the ClusterLink operator on the `clusterlink-operator` namespace
     and convert the peer certificates to secrets in the namespace where ClusterLink components will be installed.
     By default, the `clusterlink-system` namespace is used.
+    in addition it  will create a ClusterLink K8s custom resource object and deploy it to the operator.
```
elevran, line 63:
```suggestion
    in addition it will create a ClusterLink K8s custom resource object and deploy it to the operator.
```
I also think the CR type is `Instance`, not `ClusterLink`. Might be best to fully qualify it?

website/content/en/docs/tasks/operator.md
```diff
@@ -23,7 +23,7 @@ If they are not, use the flag `--path <path>` to reference the directory where c
 The command deploys the ClusterLink operator in the `clusterlink-operator` namespace and converts
 the peer certificates to secrets in the `clusterlink-system` namespace, where ClusterLink components will be installed.
 By default, these components are deployed in the `clusterlink-system` namespace.
-The `--autostart` option will create a ClusterLink K8s custom resource object and deploy it to the operator.
+In addition, the command will create a ClusterLink K8s custom resource object and deploy it to the operator.
```
elevran, line 26: ditto regarding CR type

elevran (approved):

website/content/en/docs/concepts/peers.md
kfirtoledo: I specify in more detail in the deployment operator section.

PR #512, opened 2024-04-10
hack/install-devtools.sh: Update versions

orozery:


elevran (approved):
Please make the corresponding changes in `.devcontainer/dev/` files as well

PR #523, opened 2024-04-15
controlplane/peer/client: Issue requests in parallel

orozery:
This PR changes the client to issue requests in parallel for peers who have multiple gateways.

### DIFF
--- a/pkg/controlplane/peer/client.go
+++ b/pkg/controlplane/peer/client.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"sync"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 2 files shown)


pkg/controlplane/peer/client.go
```diff
@@ -44,24 +45,60 @@ type RemoteServerAuthorizationResponse struct {
...
 
-// authorize a request for accessing a peer exported service, yielding an access token.
-func (c *Client) Authorize(req *api.AuthorizationRequest) (*RemoteServerAuthorizationResponse, error) {
-	body, err := json.Marshal(req)
-	if err != nil {
-		return nil, fmt.Errorf("unable to serialize authorization request: %w", err)
+func (c *Client) getResponse(
+	getRespFunc func(client *jsonapi.Client) (*jsonapi.Response, error),
+) (*jsonapi.Response, error) {
+	if len(c.clients) == 1 {
```
kfirtoledo, line 51: Why do you need this case and can't do iteration on 1 client?

orozery: This is the common case, and the purpose is to make it more simple since you don't need go-routines and the multi-threaded syncing in this case.

kfirtoledo (approved):

pkg/controlplane/peer/client.go
```diff
@@ -44,24 +45,60 @@ type RemoteServerAuthorizationResponse struct {
 	AccessToken string
 }
 
-// authorize a request for accessing a peer exported service, yielding an access token.
-func (c *Client) Authorize(req *api.AuthorizationRequest) (*RemoteServerAuthorizationResponse, error) {
-	body, err := json.Marshal(req)
-	if err != nil {
-		return nil, fmt.Errorf("unable to serialize authorization request: %w", err)
+func (c *Client) getResponse(
```
kfirtoledo, line 48: Can you write a comment for the function,  what it returns in case of multiple gateways (in the function description or near the code)

kfirtoledo (approved):

PR #538, opened 2024-04-24
Enable fetching all tags in actions/checkout

elevran:
Makefile uses `git describe --tags` for setting `VERSION_FLAG` during build targets. This results in an error in logs: `fatal: No names found, cannot describe anything`

- enable fetching tags in github checkout action (`fetch_tags: true`)
- avoid fetching full history (`fetch_depth: 0`) just for getting tags


.github/workflows/pr-check.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       - name: Check out repository code
         uses: actions/checkout@v4
         with:
-          fetch-depth: '0'
```
elevran, line 16: This deletion is the only part I wasn't sure about - think it is safe to replace fetch entire history with (smaller) fetch of all tags.
There is no way to fetch just the latest tag (which would have been sufficient)

kfirtoledo (approved):

PR #541, opened 2024-04-25
.github/workflows: Add release support to linux/arm64 platform

kfirtoledo:
Add release support to linux/arm64 platform.


elevran (approved):

.github/workflows/release.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       run: make push-image IMAGE_VERSION=${{ github.ref_name }}
     - name: Build and compress binaries
       run: |
-        for pair in "linux:amd64" "darwin:amd64" "darwin:arm64"; do
+        for pair in "linux:amd64" "linux:arm64" "darwin:amd64" "darwin:arm64"; do
```
elevran, line 35: you can have two `for` loops ( {linux,darwin} x {amd64,arm64} ) instead of having to parse `os`:`arch` in the next line

kfirtoledo: For now, I prefer to leave it like this, for case we want support arch only in one os

PR #564, opened 2024-05-06
Create support for arm64 architecture for docker images.

kfirtoledo:
Create support for arm64 architecture for docker images:
1) Update the GitHub action.
2) Update the makefile to compile images according to the architecture
3) Update the docker file to use the correct binary file.
4) Update the push image command to use the --platform flag for multi-architecture Docker image compilation.
5) Update the iperf3 demo to support arm64 architecture.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,8 +14,6 @@ jobs:
     steps:
     - name: checkout
       uses: actions/checkout@v4
-      with:
-        fetch-tags: true
     - name: Set up Go
       uses: actions/setup-go@v5
       with:
@@ -26,12 +24,15 @@ jobs:
         registry: ghcr.io
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
-    - name: Build images
-      run: make docker-build
-    - name: Tag and push ClusterLink images with tag 'latest'
-      run: make push-image
-    - name: Tag and push ClusterLink images latest with tag ${{ github.ref_name }}
-      run: make push-image IMAGE_VERSION=${{ github.ref_name }}
+    - name: Build binaries for amd64
+      run: make build
+    - name: Build binaries for arm64
+      run: GOARCH=arm64 make build
+    - name: Tag and push ClusterLink images with tag 'latest' and ${{ github.ref_name }}
+      run: |
+        docker buildx create --use --driver docker-container
+        PLATFORMS=linux/amd64,linux/arm64 make push-image
+        PLATFORMS=linux/amd64,linux/arm64 make push-image IMAGE_VERSION=${{ github.ref_name }}
     - name: Build and compress binaries
       run: |
         for pair in "linux:amd64" "linux:arm64" "darwin:amd64" "darwin:arm64"; do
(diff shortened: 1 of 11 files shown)


demos/iperf3/testdata/manifests/iperf3-client/iperf3-client.yaml
```diff
@@ -24,7 +24,7 @@ spec:
       containers:
       - name: iperf3-client
         #image: networkstatic/iperf3
-        image: mlabbe/iperf3
+        image: taoyou/iperf3-alpine
```
elevran, line 27: lighter iperf image, I suppose? Or does it support multi-arch which `mlabbe` does not?

demos/utils/cloud.py
```diff
@@ -68,6 +72,8 @@ def startCluster(self, testOutputFolder, logLevel="info",dataplane="envoy"):
         self.useCluster()
         super().set_kube_config()
         self.create_peer_cert(self.name,testOutputFolder)
+        if self.platform =="gcp" and self.machineType == "arm64": # Allow to schedule pods in arm64 on gcp
+            runcmd("kubectl taint nodes ` kubectl get nodes -o wide | awk '{if (NR>1) print $1}'` kubernetes.io/arch=arm64:NoSchedule-")
```
elevran, line 76: I understand it is needed but I don't understand why GCP creates nodes with `NoSchedule` in the first place. Seems odd...

elevran (approved):
lgtm, assuming the `fetch-tags: true` was dropped intentionally. If not - please fix before merging.

Makefile
```diff
@@ -3,7 +3,8 @@ SHELL=/bin/bash
 IMAGE_VERSION ?= latest
 IMAGE_ORG ?= clusterlink-net
 IMAGE_BASE ?= ghcr.io/$(IMAGE_ORG)
-
+PLATFORMS ?= linux/amd64
```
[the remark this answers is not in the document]
kfirtoledo: NO, GOOS/GOARCH  it is for the go compilation and PLATFORMS is for docker build

demos/utils/cloud.py
kfirtoledo: This problem, I saw happen to other customers.

demos/iperf3/testdata/manifests/iperf3-client/iperf3-client.yaml
kfirtoledo: Yes, It also compiles to arm64

.github/workflows/release.yml
```diff
@@ -14,8 +14,6 @@ jobs:
     steps:
     - name: checkout
       uses: actions/checkout@v4
-      with:
```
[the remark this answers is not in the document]
kfirtoledo: It caused me an error when I used it (the release wasn't tested- because this merged after the last release)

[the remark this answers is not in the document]
elevran: what was the error? 
I think we need the tags for the LD_FLAGS, otherwise it reports an error for `git describe`.
Per https://github.com/actions/checkout/issues/701#issuecomment-1139627817, an alternative would be to add another step right below the checkout that does `git fetch --tags origin`

elevran (approved):

PR #566, opened 2024-05-08
bootstrap/cert: Add an option to read only the CA certificate

kfirtoledo:
Allow to the  `deploy peer` command to read only the ca certificate for clusterlink deployment.


kfirtoledo:
replace with https://github.com/clusterlink-net/clusterlink/pull/567

PR #567, opened 2024-05-08
clusterlink deploy peer: Remove CA private keys requirement

### LINKED ISSUE #565: [BUG]: fabric key file needed for cluster deployment
**Describe the bug**
When setting up a new clusterlink fabric, the generated fabric key file is needed to deploy clusterlink on the clusters through ```clusterlink deploy```. This should instead only need the certificate file.

**To Reproduce**
Is the issue is reproducible? Always or only under certain conditions?

1. Create a new fabric.
2. Move away only the fabric private key, creating e.g.:
└── default_fabric
    ├── cert.pem
    └── cluster1
        ├── cert.pem
        ├── controlplane
        │   ├── cert.pem
        │   └── key.pem
        ├── dataplane
        │   ├── cert.pem
        │   └── key.pem
        ├── gwctl
        │   ├── cert.pem
        │   └── key.pem
        └── key.pem
3. Execute the deployment command
```
./clusterlink deploy peer --name cluster1 --ingress=NodePort --ingress-port=30443
```
4. Receive error:
```
Error: open default_fabric/key.pem: no such file or directory
```

**Expected behavior**
The ```clusterlink deploy peer``` command should only need the fabric certificate file 'default_fabric/cert.pem' for the deployment, not the fabric private key. Otherwise, this opens security concerns when creating fabrics across multiple clusters belonging to multiple parties.

**Additional context**
I am unsure where in the code base this is  Encountered on currently latest version, commit number 878d5e4eee06e6d5bb7f597bcdf120bdbe2654d4

If possible, please confirm:

- [ ] This is something I can **debug and identify the location where the problem exists**
- [ ] If the bug is confirmed, **I would be willing to submit a PR**. Bug fixes and documentation fixes are welcome.

orozery:
This PR fixes the clusterlink deploy peer command which unnecessarily required reading the fabric and peer private keys.

Fixes #565

### DIFF
--- a/pkg/bootstrap/cert.go
+++ b/pkg/bootstrap/cert.go
@@ -126,17 +126,20 @@ func CertificateFromRaw(rawCert, rawKey []byte) (*Certificate, error) {
 }
 
 // ReadCertificates read certificate and key from folder.
-func ReadCertificates(dir string) (*Certificate, error) {
+func ReadCertificates(dir string, withKey bool) (*Certificate, error) {
 	// Read certificate
 	rawCert, err := os.ReadFile(filepath.Join(dir, config.CertificateFileName))
 	if err != nil {
 		return nil, err
 	}
 
-	// Read key
-	rawFabricKey, err := os.ReadFile(filepath.Join(dir, config.PrivateKeyFileName))
-	if err != nil {
-		return nil, err
+	var rawFabricKey []byte
+	if withKey {
+		// Read key
+		rawFabricKey, err = os.ReadFile(filepath.Join(dir, config.PrivateKeyFileName))
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	cert, err := CertificateFromRaw(rawCert, rawFabricKey)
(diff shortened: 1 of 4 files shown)


pkg/bootstrap/cert.go
```diff
@@ -126,17 +126,20 @@ func CertificateFromRaw(rawCert, rawKey []byte) (*Certificate, error) {
...
 	}
 
-	// Read key
-	rawFabricKey, err := os.ReadFile(filepath.Join(dir, config.PrivateKeyFileName))
-	if err != nil {
-		return nil, err
+	var rawFabricKey []byte
+	if withKey {
+		// Read key
+		rawFabricKey, err = os.ReadFile(filepath.Join(dir, config.PrivateKeyFileName))
```
kfirtoledo, line 139: Not sure if OS.ReadFile can return nil without an error, if it can maybe it is better to propagate also the boolean "withkey" to handle this case.

kfirtoledo (approved):

PR #569, opened 2024-05-08
Update _index.md

welisheva:


michalmalka (commented):
LGTM but again i would change this in main... (it will go to next doc version 0.3)

elevran (approved):

elevran (changes requested):
Change should be done on docs/main. Only critical changes need to be "back ported" to v0.2 or v0.1

PR #571, opened 2024-05-08
Update peers.md

welisheva:
Changed a few minor things including hyphens, articles, abbreviations, fonts, subject/verb agreement, made links descriptive, etc. --

Here are a few other comments that I didn't implement because someone needs to think about them:

	Initializing a new peer:
•	Under info (in orange) – this sentence doesn’t make sense? “You will need the CA certificate (but not the CA private key) and the peer certificate and private in the next step.” Did you mean private key, the second time it’s mentioned? Certificate and private are mentioned twice and seem contradictory or redundant? Could we either take out this sentence or clarify what it meant? •	When I read this ” They can be provided out of band (e.g., over email) to the peer administrator.” it comes to mind that I would caution against sending something that is supposed to be private over email. What is the best practice? Perhaps give an explanation? 	Deploy ClusterLink to a new peer:This sentence is confusing: “Before proceeding, ensure that the CA certificate (the CA private key is not needed), and the peer certificate and key files which were created in the previous step are in the current working directory.” Reason for the confusion: No need for a private key, but yes need for key files? Public keys? If so is the word “public” worth specifying? 	Add or Remove Peers:
•	Again “out of band” use of email communication encouraged, this time a future way of handling it hinted at, consider again if this is really what you want to advise for now.

### DIFF
--- a/website/content/en/docs/v0.2/concepts/peers.md
+++ b/website/content/en/docs/v0.2/concepts/peers.md
@@ -6,23 +6,23 @@ weight: 20
 
 A *Peer* represents a location, such as a Kubernetes cluster, participating in a
  [fabric][concept-fabric]. Each peer may host one or more [services][concept-service]
- it wishes to share with other peers. A peer is managed by a peer administrator,
+ that it wishes to share with other peers. A peer is managed by a peer administrator,
  which is responsible for running the ClusterLink control and data planes. The
  administrator will typically deploy the ClusterLink components by configuring
- the [deployment CR][operator-cr]. They may also wish to provide
- (often) coarse-grained access policies in accordance with high level corporate
+ the [deployment Custom Resource (CR)][operator-cr]. The administrator may also wish
+ to provide coarse-grained access policies (and often do) in accordance with high level corporate
  policies (e.g., "production peers should only communicate with other production peers").
 
 Once a peer has been added to a fabric, it can communicate with any other peer
  belonging to it. All configuration relating to service sharing (e.g., the exporting
  and importing of services, and the setting of fine grained application policies) can be
  done with lowered privileges (e.g., by users, such as application owners). Remote peers are
- represented by the Peer Custom Resource Definition (CRD). Each Peer CR instance
+ represented by peer Custom Resource Definition (CRDs). Each Peer CR instance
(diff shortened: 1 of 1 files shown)


michalmalka (commented):
LGTM - please move changess to Main

elevran:
> 
> Initializing a new peer: 
> • Under info (in orange) – this sentence doesn’t make sense? “You will need the CA certificate (but not the CA private key) and the peer certificate and private in the next step.” Did you mean private key, the second time it’s mentioned? Certificate and private are mentioned twice and seem contradictory or redundant? Could we either take out this sentence or clarify what it meant? 

It's missing `key` in the second part. There are two pairs of certificate (the public key) and private key: one for the "fabric" (the CA) and one for the peer. You need CA certificate and both peer certificate and peer private key.

> • When I read this ” They can be provided out of band (e.g., over email) to the peer administrator.” it comes to mind that I would caution against sending something that is supposed to be private over email. What is the best practice? Perhaps give an explanation?

I'm not sure there are really good solutions or best practices for sharing secret information remotely. One could send it encrypted and provide the decryption key over the phone or any other scheme. I would leave it as is for now.
With central management (on our roadmap) this would be resolved as the information is retrieved securely from the management plane.

>  Deploy ClusterLink to a new peer:This sentence is confusing: “Before proceeding, ensure that the CA certificate (the CA private key is not needed), and the peer certificate and key files which were created in the previous step are in the current working directory.” Reason for the confusion: No need for a private key, but yes need for key files? Public keys? If so is the word “public” worth specifying? 

To clarify: the (1) CA certificate, (2) peer certificate and (3) peer private key are needed. The CA private key is not needed in this step. All these three are stored in files which should be available in the local working do
Happy to integrate a more clearly written sentence.

>  Add or Remove Peers: • Again “out of band” use of email communication encouraged, this time a future way of handling it hinted at, consider again if this is really what you want to advise for now.

I don't see an immediate alternative. Customer should use whatever mechanism they use today for handling similar situations.

elevran (changes requested):
Change should be done on docs/main. Only critical changes need to be "back ported" to v0.2 or v0.1
Please see responses to the comments in the PR message.

PR #591, opened 2024-05-19
copyright: update all copyright lines and remove year

### LINKED ISSUE #285: Ensure copyrights are updated for 2024
The copyright check as well as generated code (e.g., CRD) should reflect 2024 in the copyright

elevran:
According to https://hynek.me/til/copyright-years/ and linked sources (including Linux Foundation) copyright years:

- are mostly pointless in all major jurisdictions
- are already dropped by many big corporations and projects
- cause pointless churn on every January and git already keeps history for us...

the years range (`2022-present`) is kept in LICENSE.
The copyright in `.licenserc.yaml` is updated accordingly.

Fixes #285


michalmalka (changes requested):
You must delete current copyright - as it is now duplicated and not updated as expected... it is a problem of the tool that we're using probably.

elevran:
> You must delete current copyright - as it is now duplicated and not updated as expected... it is a problem of the tool that we're using probably.

Agree. I noticed that too and put the PR in draft mode... 
It is a tool limitation so resorted to using `sed` instead: 
```sh
grep -rl "Copyright 2023 The ClusterLink Authors" --exclude-dir=.git | xargs sed -i 's/Copyright 2023 The ClusterLink Authors/Copyright (c) The ClusterLink Authors/g'
```

This will be in a new PR (also need to run codegen on CRDs)

PR #604, opened 2024-05-22
Create SECURITY.md

welisheva22:
Given that we are in the alpha stage, I think this would suffice. When we are out of alpha stage, it would be valuable to revise this process as security vulnerabilities may need to be handled differently than other enhancements/bugs/comments.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,22 +1,51 @@
-# Security Policy
+# Security policy
 
-## Supported Versions
+Thank you for your interest in the security of the ClusterLink project.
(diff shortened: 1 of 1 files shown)


elevran:
Not sure we want vulnerabilities reported as normal issues. Need to research what options are available so they can be fixed before disclosed publicly (e.g., maintainers mailing list)

elevran (approved):

elevran:
Enabled private vulnerability reporting on the repo. Revised SECURITY.md text to reflect that.

PR #609, opened 2024-05-23
Send attributes of source workload to remote peer

### LINKED ISSUE #293: Sending attributes of the client workload in a connection request
When a workload is trying to connect to a remote service, the local peer must send all workload attributes to the remote peer, so the remote peer can make a decision whether access is permitted or not.

zivnevo:
Fixes #293

### DIFF
--- a/pkg/controlplane/api/authz.go
+++ b/pkg/controlplane/api/authz.go
@@ -13,7 +13,10 @@
 
 package api
 
-import "github.com/lestrrat-go/jwx/jwa"
+import (
+	"github.com/clusterlink-net/clusterlink/pkg/controlplane/authz/connectivitypdp"
+	"github.com/lestrrat-go/jwx/jwa"
+)
 
 const (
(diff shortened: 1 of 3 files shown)


elevran:
where does the collection of attributes done?

zivnevo:
> where does the collection of attributes done?

Currently it is done on the peer at the client-workload side when authorizing the egress request (`Manager::authorizeEgress`).
We may decide to move it to another function/struct when there is enough "beef" around it.
The idea behind this PR is that both peers will use exactly the same set of attributes for the source workload.

orozery:
@elevran Recall your suggestion of switching tunnel-establishment

orozery:
Sorry, closed by mistake. Re-opened.

elevran (approved):

elevran:
so this just makes sure we send/use the attributes and does not affect which are collect, right?

zivnevo:
> so this just makes sure we send/use the attributes and does not affect which are collect, right?

right

Issue #599, opened 2024-05-21
Compatibility for Mac OS


welisheva22:
Hi- while going through the documentation I decided to take the tutorials to make sure I had a good understanding of the concepts, before suggesting images to add. I ran into the following issue.

Allow me to describe my experience getting started.

The documention said to install kind. The kind documentation says to also install other technologies. At first I wasn't sure whether to install docker or another technology. After trying podman and nerdctl via lima, I ended up choosing docker because the kind technology was actually looking for the word docker when I used it. There was finally some success there! I ran a kind command and got several smiley faces that I was hoping would appear.

I then went back to the basic users getting started docs for ClusterLink and ran the curl command to install it. It didn't take long and said it was downloading a file with .tar.gz in it -- but I didn't see that file anywhere on the machine. I tried typing clusterlink --version and it didn't know what the clusterlink command meant.

I went fishing for it. I found it online here:
https://github.com/clusterlink-net/clusterlink/releases/download/v0.2.1/clusterlink-darwin-amd64.tar.gz

After I downloaded and ran the tar.gz file I was excited to see (via Finder) the word "clusterlink" in what looked like a runnable file on my computer! I went into .zprofiles and added that to the path. Then I typed clusterlink --version but it gave me the attached message. My computer seemed afraid to run it for some reason!
![error message checking clusterlink version](https://github.com/clusterlink-net/clusterlink/assets/169834950/efdf5558-6826-4c32-8201-3d0e14e65fb4)

I researched the error message on the web and found this link.
https://www.lifewire.com/fix-developer-cannot-be-verified-error-5183898 

I followed the instructions to override the caution by double clicking on the two files in the new clusterlink folder and manually authorizing their use. Finally, it was able to tell me which version of ClusterLink was installed via the clusterlink --version command. Success!

So the issue is:

1) Let's remember to get on the approved list of apple developers so that others don't need to go through this hassle

2) Until we are on their list would it be advisable to teach about the workaround in the documentation, or would that just open up a can of worms?

It can certainly be dissuading to users to contribute to the open source code base, when the computer advises against it due to malware suspicion, just because we're not an approved Apple developer yet. I look forward to our efforts to change that someday.

Thanks!

welisheva22:
I noticed in the file
[clusterlink](https://github.com/clusterlink-net/clusterlink/tree/main)/[hack](https://github.com/clusterlink-net/clusterlink/tree/main/hack)
/install_clusterlink.sh

there is code:
[...]

PR #614, opened 2024-05-23
Update install_clusterlink.sh

welisheva22:
documentation
Changed spelling in comment (added an e in determine both times it appeared)

code
Added prefix CL_ to variable CL_ARCH


elevran (approved):

elevran:
Replaces #610 
Ref #599

PR #618, opened 2024-05-27
Add import and export labels as destination attributes

### LINKED ISSUE #292: Setting service attributes, to be used in access policies
Probably set in the Export CRD.
How do we publish these attributes to all remote peers that import the service?

zivnevo:
Fixes #292 

Also, enriching `util.Service` to have labels, so we can better test this new addition.


pkg/controlplane/authz/manager.go
```diff
@@ -230,11 +236,9 @@ func (m *Manager) authorizeEgress(ctx context.Context, req *egressAuthorizationR
 			}
 		}
 
-		dstAttributes := connectivitypdp.WorkloadAttrs{
-			ServiceNameLabel:      imp.Name,
-			ServiceNamespaceLabel: imp.Namespace,
-			GatewayNameLabel:      importSource.Peer,
-		}
+		dstAttributes[ServiceNamespaceLabel] = importSource.ExportNamespace
```
orozery, line 239: Why is `ServiceNameLabel` copied from the import, and on the other hand `ServiceNamespaceLabel` is copied from the export?

zivnevo: It's a decision we have to take - how do you call the remote service locally?
The import name? The export name? Have two attributes to give the user the full flexibility?

orozery: I think it's confusing to mix both, so either give full flexibility, or be consistent (use just import or just export).

zivnevo: Changed to use `importSource.ExportName` as the name of the remote service.
This way, the PDPs at both ends of the connection should always agree on the value for this attribute.

orozery (approved):

PR #620, opened 2024-05-28
website: Extend the explanation in the iperf3 tutorial

kfirtoledo:
1. Add an explanation about the certificate creation needs to use the same fabric CA.
1. Add a check for peer status to verify peer connectivity.
Preview:
https://deploy-preview-620--clusterlink-net.netlify.app/docs/main/tutorials/iperf/#set-up-peers


website/static/files/tutorials/deploy-clusterlink.md
```diff
@@ -14,6 +14,10 @@
     clusterlink create peer-cert --name server
     ```
 
+    All peer certificates (client and server) should be created from the same Fabric CA files.
```
elevran, line 17:
```suggestion
    All peer certificates (i.e., for the `client` and `server` clusters, in this tutorial) should be created from the same fabric CA files.
```

elevran: Since fabric is lower case in line 2, I changed to `fabric` here too.

website/static/files/tutorials/deploy-clusterlink.md
```diff
@@ -14,6 +14,10 @@
...
+    In this tutorial, we assume the server has access to the Fabric certificate created in the `default_fabric` folder.
+    In case the server doesn't have access, the Fabric certificate should be copied from the client to the server.
```
elevran, line 19:
```suggestion
    In this tutorial, we assume the `server` cluster creation has access to the fabric certificate stored in the
     `default_fabric` folder. If it doesn't, the fabric certificate should be copied from the `client` to the `server`.
```

elevran (approved):

PR #623, opened 2024-05-29
CLI: Enhancement for creation of ClusterLink components' certificates

kfirtoledo:
Create the ClusterLink components' certificates on the `deploy peer`  instead of in the `create peer-cert` command.


elevran (approved):
- all: deploy operator, deploy CR. No files needed locally afterwards
- operator: deploy operator, deploy secrets, optionally leave a sample CR file
- none: all k8s resources in files (without operator and CR of course). No need to deploy CRD and operator RBAC

PR #627, opened 2024-06-03
Setup Go with the Go version in go.mod

zivnevo:


zivnevo:
Currently, some required checks appear as yellow circles, as check names have changes because we no longer use matrix testing.
Once the PR is approved, I'll change the check names on the branch-protection rule.

elevran (approved):

PR #629, opened 2024-06-03
cmd/cl-controlplane: Support dynamic peer certificates

orozery:
This PR changes the controlplane to watch the peer certificate files for changes, in order to support dynamic peer certificates.

Note that when switching TLS certificates, existing connections are still valid.
The client-side controlplane (the importer) terminates those controlplane connections when switching certificates.
However, if only the server switches certificates, the server will continue to allow authz connections, which will be dropped only at the tunnel establishment phase (where a new connection is used).

### DIFF
--- a/pkg/controlplane/peer/certs.go
+++ b/pkg/controlplane/peer/certs.go
@@ -1,0 +1,166 @@
+// Copyright (c) The ClusterLink Authors.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package peer
+
+import (
+	"errors"
+	"fmt"
+	"path"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
+	"github.com/sirupsen/logrus"
+
+	"github.com/clusterlink-net/clusterlink/pkg/util/tls"
+)
+
+// CertsConsumer represents a consumer of peer TLS certificates.
+type CertsConsumer interface {
+	SetPeerCertificates(parsedCertData *tls.ParsedCertData, rawCertData *tls.RawCertData) error
+}
+
+// parseError represents an error in parsing the peer certificates.
+type parseError struct {
+	err error
+}
+
+func (e parseError) Error() string {
+	return e.err.Error()
+}
+
+// CertsWatcher watches certificate updates.
+type CertsWatcher struct {
+	caPath   string
+	certPath string
+	keyPath  string
+
+	stopCh    chan struct{}
+	consumers []CertsConsumer
+
+	logger *logrus.Entry
+}
+
+// Name of the watcher.
+func (w *CertsWatcher) Name() string {
+	return "certs-watcher"
+}
+
+// AddConsumer adds a new peer certificates consumer.
+// This function is not thread-safe.
+func (w *CertsWatcher) AddConsumer(consumer CertsConsumer) {
+	w.consumers = append(w.consumers, consumer)
+}
+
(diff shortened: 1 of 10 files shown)


elevran:
> The client-side controlplane (the importer) terminates those controlplane connections when switching certificates.

Am I reading this correctly: 
When the client side control plane has a new certificate, it will re-establish (and therefore reauthenticate) *control connections* to remote peers.

This does not refer to existing workload-to-workload connections which are kept intact. 
Correct?

orozery:
> > The client-side controlplane (the importer) terminates those controlplane connections when switching certificates.
> 
> Am I reading this correctly: When the client side control plane has a new certificate, it will re-establish (and therefore reauthenticate) _control connections_ to remote peers.
> 
> This does not refer to existing workload-to-workload connections which are kept intact. Correct?

Correct

pkg/controlplane/peer/certs.go
elevran, line 1: nit: this seems independent of peer. Consider moving it to its own package 
(which also means you can leverage the package name instead of setting `peerCertWatcher` and the likes in callers)

pkg/controlplane/peer/certs.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// Name of the watcher.
+func (w *CertsWatcher) Name() string {
+	return "certs-watcher"
+}
+
+// AddConsumer adds a new peer certificates consumer.
+func (w *CertsWatcher) AddConsumer(consumer CertsConsumer) {
+	w.consumers = append(w.consumers, consumer)
```
elevran, line 61: single goroutine access is assumed. Fine if it is internal, may want to protect (or at least document) if can be used by any part of the code.

pkg/controlplane/peer/certs.go
elevran, line 1: Consider if can use https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/certwatcher instead of implementing it in CL

pkg/controlplane/peer/certs.go
```diff
@@ -0,0 +1,165 @@
...
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
+
+	certsModified := false
+	for {
+		select {
+		case <-w.stopCh:
+			return nil
+		case event := <-watcher.Events:
+			w.logger.Infof("Event: %v", event)
```
elevran, line 117: nit: useful in debug, night not be needed in production. I don't expect many notifications here but maybe it should be debug level and the update function would log the change

pkg/controlplane/peer/certs.go
```diff
@@ -0,0 +1,165 @@
...
+			certsModified = true
+		case err := <-watcher.Errors:
+			w.logger.Infof("Error: %v", err)
+			return err
+		case <-ticker.C:
```
elevran, line 122: nit: an alternative to reloading on timer is to look at how configmap changes are detected and the symlinks followed to determine change. See for example in [viper](https://github.com/spf13/viper/commit/e0f7631cf3ac7e7530949c7e154855076b0a4c17).
I'm ok with it as is, but it may be a useful package in which case you may want to have that as a more complete implementation, not relying on timers.

elevran (approved):
Please consider if use this code as-is, change to an existing certwatcher (e.g., from controller runtime), or move the code into its own package

pkg/controlplane/peer/certs.go
orozery: This was my first thought, to make this code more generic and place it in `pkg/util`.
However, currently there's no other catcher of this, and placing it in `peer` allowed me to tweak it for the specific use-case (watch dirs instead of files, ignore operation type, batch events using a ticktimer, ignore certain types of callback errors (`parseError`)).

pkg/controlplane/peer/certs.go
orozery: Sorry, I don't follow.
Let's talk about it in our next meeting.

elevran: currently: mark the file change event, every second (on timer) reload if there was a change event.
viper: on change event - follow symlinks to detect if the actual file content changed and reload it.

Viper does this for configmap mounted as a volume which is similar to how secrets are mounted.
I was just saying that you *could* look at the code that actually examines the file that holds the data instead of reloading based on timer. It is not critical

PR #634, opened 2024-06-05
controlplane: Change listen port number

kfirtoledo:
Update the controlplane port to be higher than 1024 (below does not allow for non-root users in OpenShift clusters).

### DIFF
--- a/pkg/controlplane/api/servername.go
+++ b/pkg/controlplane/api/servername.go
@@ -15,5 +15,5 @@ package api
 
 const (
 	// ListenPort is the port used by the dataplane to access the controlplane.
-	ListenPort = 444
+	ListenPort = 4444
 )


pkg/controlplane/api/servername.go
```diff
@@ -15,5 +15,5 @@ package api
 
 const (
 	// ListenPort is the port used by the dataplane to access the controlplane.
-	ListenPort = 444
+	ListenPort = 4444
```
elevran, line 18: fyi: there is an IANA registry of ports, better pick an unassigned one (e.g., https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?&page=88).
Also next time anyone wants a different port - lets just make it configurable via CLI and use this value as default ;-)

elevran (approved):

orozery (approved):

PR #635, opened 2024-06-05
installation: Create a temporary directory for ClusterLink CLI installation

kfirtoledo:
Create a temporary directory for downloading and installing clusterlink CLI.


hack/install_clusterlink.sh
```diff
@@ -1,4 +1,22 @@
 #!/bin/sh
+# Copyright 2024 The ClusterLink Authors.
```
elevran, line 2: There should not be any copyright year in the file. This will fail the license check as it does not meet the expected text.
See `.licenserc.yaml` configuration at the top level

hack/install_clusterlink.sh
```diff
@@ -42,14 +60,12 @@ if ! curl -o /dev/null -sIf "$url"; then
 fi
 
 # Open the tar file.
+current_path=$(mktemp -d "$(pwd)/clusterlink.XXXXXX")
```
elevran, line 63: nit: download_path?

elevran (approved):
Please fix the copyright before comitting

PR #638, opened 2024-06-06
CLI: Add zero bytes to empty certificates

kfirtoledo:
Add zero bytes to empty certificates that are used for cleanup by ClusterLink CLI.

### DIFF
--- a/pkg/bootstrap/platform/k8s.go
+++ b/pkg/bootstrap/platform/k8s.go
@@ -375,20 +375,16 @@ func K8SClusterLinkInstanceConfig(config *Config, name string) ([]byte, error) {
 // K8SEmptyCertificateConfig returns Kubernetes empty secrets for the control plane and data plane,
 // used for deleting the secrets.
 func K8SEmptyCertificateConfig(config *Config) ([]byte, error) {
-	// Create a base64-encoded string of zeros
-	zeroBytes := make([]byte, 64)
-	base64Zeros := base64.StdEncoding.EncodeToString(zeroBytes)
-
 	args := map[string]interface{}{
 		"Namespace":        config.Namespace,
-		"ca":               base64Zeros,
-		"controlplaneCert": base64Zeros,
-		"controlplaneKey":  base64Zeros,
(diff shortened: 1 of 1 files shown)


kfirtoledo:
In the cleanup phase, we delete all the secrets.
For deletion is enough to know the name of the secret object, but there is still validation that runs on the secrets.
 This fix ignores the error validation of the empty secrets.

elevran:
> In the cleanup phase, we delete all the secrets. For deletion is enough to know the name of the secret object, but there is still validation that runs on the secrets. This fix ignores the error validation of the empty secrets.

I don't understand why there's validation on a secret that's deleted.
Who is validating? ClusterLink, Envoy, K8s?
The secrets are mounted into Deployment volumes, might this be the problem? Perhaps, you can't delete a secret that is mounted and that's why it needs to hold a value?
If that's the case, then we need to make sure we delete the objects in reverse order of creation: secrets were created before Pods so deletion should be first CL Pods/Deployments and then secrets after they're no longer needed.

kfirtoledo:
This code was accidentally deleted in the TLS refactoring PR https://github.com/clusterlink-net/clusterlink/pull/619/files#:~:text=%22fabricCA%22%3A%20%20%20%20%20%20%20%20%20%22%22,dataplaneKey%22%3A%20%20%20%20%20%22%22%2C
Without this code, the secret looks:
---
apiVersion: v1
kind: Secret
metadata:
  name: cl-ca
  namespace: <no value>
data:
  ca: <no value>
---
apiVersion: v1
kind: Secret
metadata:
  name: cl-controlplane
  namespace: <no value>
data:
  cert: <no value>
  key: <no value>
---
Those are not valid certificates.
I can create a secret object for each one and delete it, But I prefer to reuse the code and delete it in the same way as the secrets were created.
The validation is done as part of the package `sigs.k8s.io/e2e-framework/klient/k8s/resources`, because this type of secret is not legal. 
About the order of the deletion, it doesn't matter because we use the DeleteIgnoreNotFound function.

PR #639, opened 2024-06-09
.github/workflows: Add CLI installation checks for differnt platforms 

kfirtoledo:


kfirtoledo:
Will check first in my local fork.

PR #645, opened 2024-06-13
controlplane: Add error message when hearbeat fail.

kfirtoledo:
Add logs to controlplane when a heartbeat is fail

### DIFF
--- a/pkg/controlplane/control/peer.go
+++ b/pkg/controlplane/control/peer.go
@@ -121,7 +121,8 @@ func (m *peerMonitor) Start() {
 			break
 		}
 
-		heartbeatOK := m.getClient().GetHeartbeat() == nil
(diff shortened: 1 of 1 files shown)


elevran (approved):

pkg/controlplane/control/peer.go
```diff
@@ -138,6 +139,8 @@ func (m *peerMonitor) Start() {
 		if strikeCount < threshold {
 			<-ticker.C
 			continue
+		} else if heartbeatErr != nil {
+			m.logger.Errorf("Fail to sent heartbeat to %s : %v", m.pr.Name, heartbeatErr)
```
elevran, line 143:
```suggestion
			m.logger.Errorf("Failed to send heartbeat to %s: %v", m.pr.Name, heartbeatErr)
```
Would be better English

PR #646, opened 2024-06-13
website: Fix allow-all-policy.md

kfirtoledo:
`readfile` function is not able to support `tabpane` for more than one file, so I  remove the `tabpane` from the included file


elevran (approved):
Have you reached out to the Hugo/Docsy community with a question/bug report?
We are probably not the first to have encountered failures in using shortcodes inside an included file

PR #651, opened 2024-06-17
.github/workflows: Add CLI installation checks.

kfirtoledo:
Add installation check using the CLI :
This test checks the CLI commands for creating certificates, deploying peers, and deleting them.

### DIFF
--- a/.github/workflows/pr-cli.yml
+++ b/.github/workflows/pr-cli.yml
@@ -1,0 +1,39 @@
+name: PR check - CLI Installation test
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  cli-test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+
+    steps:
+    - name: checkout
+      uses: actions/checkout@v4
+      with:
+        fetch-tags: true
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version-file: ./go.mod
+    - name: Install kind
+      uses: helm/kind-action@v1.10.0
+      with:
+        install_only: true
+    - name: Run build
+      run: make build
+    - name: Build docker images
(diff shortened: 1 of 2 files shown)


tests/cli/basic_test.sh
```diff
@@ -0,0 +1,80 @@
...
+
+cd $TEST_DIR
+clean_up
+
+trap clean_up_with_logs INT TERM EXIT
+
+cd $TEST_DIR
+test_k8s
+
+echo OK
```
elevran, line 80: nit: missing EoL

tests/cli/basic_test.sh
```diff
@@ -0,0 +1,80 @@
...
+        echo "Error: Timeout occurred while waiting for cl-controlplane deployment"
+        exit 1
+    fi
+
+    if ! timeout 30 sh -c 'until kubectl rollout status deployment cl-dataplane -n clusterlink-system; do sleep 0.1; done > /dev/null 2>&1'; then
+        echo "Error: Timeout occurred while waiting for cl-dataplane deployment"
+        exit 1
+    fi
+
+  # Selete clusterlink objects
```
elevran, line 67:
```suggestion
  # Delete clusterlink objects
```

.github/workflows/pr-cli.yml
```diff
@@ -0,0 +1,39 @@
+name: PR check - CLI Installation test
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  cli-test:
+    runs-on: ${{ matrix.os }}
```
elevran, line 10: the PR says "for different platforms" but it's only running on ubuntu-latest (IIUC the use of `matrix.os` from below)

elevran (approved):

.github/workflows/pr-cli.yml
kfirtoledo: Yes, you are right, I will fix the description.

PR #654, opened 2024-06-18
controlplane: Store JWK keys in k8s secret

orozery:
Today, the JWK is initialized by the controlplane when loading, and stored in-memory.
To support multiple controlpanes with the same JWK key, we change the controlplane authz manager to ready the JWK key from a secret, instead of generating it.manager
The secret is created by the (soon to be leader elected) control manager.

Note: rotating JWK keys is not yet supported. Planning to open a feature issue after this PR is merged.


pkg/operator/controller/instance_controller.go
```diff
@@ -65,6 +65,7 @@ type InstanceReconciler struct {
 // +kubebuilder:rbac:groups=clusterlink.net,resources=instances,verbs=list;get;watch;update;patch
 // +kubebuilder:rbac:groups=clusterlink.net,resources=instances/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=clusterlink.net,resources=instances/finalizers,verbs=update
+// +kubebuilder:rbac:groups="",resources=secrets,verbs=list;get;watch;create;update
```
elevran, line 68: nit: is there a need for Cluster scope or can we create a Namespaced role for reading just the secret in our namespace?
you can also further limit using resourceNames in the kubebuilder annotation

elevran (approved):

pkg/operator/controller/instance_controller.go
orozery: Although by default the namespace is "clusterlink-system", it is customizable.
So I don't think we can use the namespace annotation.
Also the resource name is defined in a go const string. I don't see how can I refer to it in the kubebuilder comment.

PR #656, opened 2024-06-18
Add CoreDNS rewrite support for external services.

### LINKED ISSUE #26: Improve DNS support
In TLS communication, the client accessing a service via ClusterLink gateways requires that the local name resolved appears in the certificate presented by the server.

For k8s services, it means we need to keep the source and destination DNS the same (i.e., maintain the name and namespace between clusters). This may require cluster wide privileges (i.e., create a service in a different namespace than the one the gateway runs in).
This is even more apparent when importing a service that is external to a cluster and uses TLS, The client needs to resolve the name as it appears in the certificate (e.g., api.example.com) to the local gateway. Note that a secondary local domain (e.g., "multicluster.local" or "foo.io") would not work as it too would not be present in the cloud service's certificate.

Kubernetes allows this via a few mechanisms:
- manually (or via Admission controller) the DNS configuration of the client Pod to use a special DNS server that can do the overrides
- change the CoreDNS configuration to rewrite/resolve external names in a special way

For reference on changing Pod DNS see [here](https://itnext.io/kubernetes-hacks-and-tricks-5-pod-dns-policy-and-dns-config-options-c176f4a581c3). For examples manipulating CoreDNS see [here](https://coredns.io/2017/05/08/custom-dns-entries-for-kubernetes/) and [here](https://coredns.io/2017/06/08/how-queries-are-processed-in-coredns/)

In addition, we would need to extend the service import (or binding) to have an optional DNS alias name that includes the expected client DNS entry.
The hope is that the above also can be used to resolve issues with protocols that seed client with bootstrap servers which are in turn used to get the full list of servers (e.g., Kafka does that). In that case, we would need to alias the full list of servers that can be discovered by the client

aviweit:
This PR deals with updating CoreDNS configuration to rewrite/resolve external names as described in issue #26.

Fixes #26


aviweit:
End to end tests were made with two EKS AWS clusters with three Import CRs:

1. internal nginx service
2. external s3 using vpe dns name
3. external lambda using vpe dns name

these CRs had been added and removed during the tests - to ensure coredns configmap is properly updated.

It was also tested with two local kind clusters ensuring that coredns is properly updated.

Thanks.

config/operator/rbac/role.yaml
```diff
@@ -4,6 +4,15 @@ kind: ClusterRole
 metadata:
   name: cl-operator-manager-role
 rules:
+- apiGroups:
```
elevran, line 7: nit: could this be made a namespaced role solely for the CoreDNS config map and the k8s-system namespace? can it be reduced further to only the specific config map?
This can be done by adding [`resourceNames`](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources)

config/crds/clusterlink.net_imports.yaml
```diff
@@ -40,6 +40,10 @@ spec:
           spec:
             description: Spec represents the attributes of the imported service.
             properties:
+              dnsName:
+                description: DnsName is an optional external dns name for this imported
```
elevran, line 44: Based on Go's convention this should be `DNSName` (should be changed in the Go struct and then generated to here).
The YAML field can still be `dnsName`

pkg/apis/clusterlink.net/v1alpha1/import.go
```diff
@@ -65,6 +65,8 @@ type ImportSpec struct {
 	// +kubebuilder:default="round-robin"
 	// LBScheme is the load-balancing scheme to use (e.g., random, static, round-robin)
 	LBScheme LBScheme `json:"lbScheme"`
+	// DnsName is an optional external dns name for this imported service
+	DnsName string `json:"dnsName,omitempty"`
```
elevran, line 69: Based on Go's convention this should be `DNSName`

```suggestion
	// DNSName is an optional external DNS name for this imported service
	DNSName string `json:"dnsName,omitempty"`
```

elevran: Also, `Alias` might be a better name since we create an alias for the Service (it has a DNS name given by the Service and k8s)

pkg/controlplane/control/manager.go
```diff
@@ -243,31 +336,96 @@ func (m *Manager) AddImport(ctx context.Context, imp *v1alpha1.Import) (err erro
...
+	return nil
 }
 
 // DeleteImport removes the listening socket of a previously imported service.
 func (m *Manager) DeleteImport(ctx context.Context, name types.NamespacedName) error {
 	m.logger.Infof("Deleting import '%s/%s'.", name.Namespace, name.Name)
 
 	// delete user service
-	errs := make([]error, 3)
+	errs := make([]error, 4)
```
elevran, line 428: nit: I realize it is not part of the new code, but I wonder if using append won't be a cleaner solution instead of assigning specific array entries. Unless, of course, there is a reliance on the specific order even when there are nil errors in the array.
Alternately, this might be better served by `errors.Join()`.
@kfirtoledo @orozery - not familiar with this code well enough to say. Please weigh in.

pkg/operator/controller/instance_controller.go
```diff
@@ -68,7 +68,8 @@ type InstanceReconciler struct {
 // +kubebuilder:rbac:groups="",resources=services;serviceaccounts,verbs=list;get;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="discovery.k8s.io",resources=endpointslices,verbs=list;get;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=nodes,verbs=list;get;watch
-// +kubebuilder:rbac:groups="",resources=pods,verbs=list;get;watch
+// +kubebuilder:rbac:groups="",resources=pods,verbs=list;get;delete;watch
```
elevran, line 71: might be cleaner to add RBAC for deployment as a whole and not for Pods (BTW, it can be constrained to the specific CoreDNS deployment by using `resourceNames` - not sure if  it needs to be a ClusterRole or can be namespaced. kubebuilder [rbac annotations](https://book.kubebuilder.io/reference/markers/rbac) can be used to generate what we want.

pkg/operator/controller/instance_controller.go
```diff
@@ -428,6 +429,13 @@ func (r *InstanceReconciler) createAccessControl(ctx context.Context, name, name
 					"get", "list", "watch", "create", "delete", "update",
 				},
 			},
+			{
+				APIGroups: []string{""},
+				Resources: []string{"configmaps"},
+				Verbs: []string{
+					"get", "list", "update", "watch",
+				},
+			},
```
elevran, line 438: please change here too if you decide to use `resourceNames`

pkg/operator/controller/instance_controller.go
```diff
@@ -438,7 +446,7 @@ func (r *InstanceReconciler) createAccessControl(ctx context.Context, name, name
 			{
 				APIGroups: []string{""},
 				Resources: []string{"pods"},
-				Verbs:     []string{"get", "list", "watch"},
+				Verbs:     []string{"get", "delete", "list", "watch"},
```
elevran, line 449: see earlier comment on pods vs deployment.
So far we did not need to delete Pods, only read access was needed for determining the pod's IP addresses.

pkg/controlplane/control/manager.go
```diff
@@ -144,6 +144,99 @@ type Manager struct {
 	logger *logrus.Entry
 }
 
+// Restart coredns pods
+func coreDnsRestart(ctx context.Context, m *Manager) error {
```
elevran, line 148: I would prefer to do this using CoreDNS deployment changes and not by deleting individual Pods, if possible.
The deployment has a specific name so you can just get it directly, instead of iterating over all pods in the namespace.
Looking to do the equivalent of ` kubectl rollout restart deployment...`

pkg/controlplane/control/manager.go
```diff
@@ -144,6 +144,99 @@ type Manager struct {
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+// Add coredns rewrite for a given external dns service
```
elevran, line 167: It may be less intrusive to use the `hosts` or `file` [plugins](https://coredns.io/plugins) so that the changes are not embedded as part of the CoreDNS configuration but are in their own file.
We could then ask an administrator to enable the CoreDNS plugin manually (if they so desire) and require lower privileges for us (we won't be mocking with the Corefile directly).

pkg/controlplane/control/manager.go
```diff
@@ -144,6 +144,99 @@ type Manager struct {
...
+	if err := m.client.Get(ctx, corednsName, &cm); err != nil {
+		if k8serrors.IsNotFound(err) {
+			m.logger.Warnf("coredns configmap not found.")
+			return nil
+		} else {
+			return err
+		}
+	}
+	if data, ok := cm.Data["Corefile"]; ok {
+		// remove trailing end-of-line
```
elevran, line 184: if you use the file / hosts plugins there might already be packages for parsing their data (I imagine the pluiin needs it anyway)

pkg/controlplane/control/manager.go
```diff
@@ -144,6 +144,99 @@ type Manager struct {
...
+		dataEol := strings.TrimSuffix(data, "\n")
```
elevran, line 185: nit: reassign into `data`?

pkg/controlplane/control/manager.go
```diff
@@ -144,6 +144,99 @@ type Manager struct {
...
+		var rewriteLine = ""
+		for i, line := range lines {
+			if strings.Contains(line, serviceFqdn) {
+				// matched line already exists
+				break
+			}
+			// ready marker is reached - matched line not found, append it here
+			if strings.Contains(line, "    ready") {
+				if strings.HasPrefix(dnsName, "*.") {
+					// wildcard dns
```
elevran, line 200:
```suggestion
				if strings.HasPrefix(dnsName, "*.") { // wildcard DNS
```

pkg/controlplane/control/manager.go
```diff
@@ -144,6 +144,99 @@ type Manager struct {
...
+					dnsName = strings.TrimPrefix(dnsName, "*")
+					dnsName = strings.ReplaceAll(dnsName, ".", "\\.")
+					dnsName = "(.*)" + dnsName
+
+					rewriteLine = fmt.Sprintf("    rewrite name regex %s %s answer auto", dnsName, serviceFqdn)
```
elevran, line 205: nit: Is a tab expected before `rewrite` keyword? If so, consider using `\t` instead of spaces

pkg/controlplane/control/manager.go
```diff
@@ -243,31 +336,96 @@ func (m *Manager) AddImport(ctx context.Context, imp *v1alpha1.Import) (err erro
...
-	return m.addImportService(ctx, imp, userService)
+	if err := m.client.Get(ctx, corednsName, &cm); err != nil {
+		if k8serrors.IsNotFound(err) {
+			m.logger.Warnf("coredns configmap not found.")
+			return nil
+		} else {
+			return err
+		}
+	}
+	if data, ok := cm.Data["Corefile"]; ok {
```
elevran, line 383: may want to move rewrite parsing into its own package/struct so that the file parsing is not split across these functions as internal implementation

elevran (changes requested):
@aviweit - thank you for the pull request and adding this feature!

Can you please 
- use uppercase `DNS` in function names and fields exposed?
- move all DNS related functions into their own file so changes to existing code files are minimized? 
- Consider if the use case is best served by `rewrite`, `hosts` or `file` CoreDNS plugin?

Thanks!

pkg/controlplane/control/manager.go
aviweit: Done

pkg/controlplane/control/manager.go
aviweit: Yes, done.

pkg/controlplane/control/manager.go
```diff
@@ -144,6 +144,98 @@ type Manager struct {
 	logger *logrus.Entry
 }
 
+// Restart coredns pods
+func coreDnsRestart(ctx context.Context, m *Manager) error {
```
kfirtoledo, line 148: Can the function be part of the manager class?

pkg/apis/clusterlink.net/v1alpha1/import.go
aviweit: Updated to Alias.

pkg/controlplane/control/manager.go
elevran: dns.go would be a more natural place for it, IMO. It is not really Manager'ish... 
Also, may want to narrow the scope of parameter passed (e.g., would a Client suffice or do you need the full Manager?)

pkg/controlplane/control/dns.go
```diff
@@ -0,0 +1,174 @@
...
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Restart coredns pods
+func coreDnsRestart(ctx context.Context, m *Manager) error {
```
elevran, line 29: Suggest passing in the logger and client directly, instead of the full manager struct.
The same should be applied to other functions in this file

elevran, line 29: make this consistent with other functions which have the form `<verb><subject>`. Also upcase `DNS`
```suggestion
func restartCoreDNS(ctx context.Context, m *Manager) error {
```

pkg/controlplane/control/manager.go
```diff
@@ -243,31 +243,40 @@ func (m *Manager) AddImport(ctx context.Context, imp *v1alpha1.Import) (err erro
...
-			Namespace: imp.Namespace,
-			Labels:    make(map[string]string),
-		},
-		Spec: v1.ServiceSpec{
-			ExternalName: fmt.Sprintf("%s.%s.svc.cluster.local", serviceName, m.namespace),
-			Type:         v1.ServiceTypeExternalName,
-		},
+	if imp.Spec.Alias != "" {
+		if err := addCoreDnsRewrite(ctx, m, &importName, imp.Spec.Alias); err != nil {
+			m.logger.Errorf("Failed to configure CoreDns: %v.", err)
```
elevran, line 266:
```suggestion
			m.logger.Errorf("failed to configure CoreDNS: %v.", err)
```

elevran: Also, if `logger.Errorf` supports `%w`  - prefer that over `%v`

pkg/controlplane/control/manager.go
aviweit: `\t` did not work although it is 4 chars. It might be that it is 4 spaces rather than tab. I reverted this back.

pkg/controlplane/control/manager.go
aviweit: I get compiler warning: `(*github.com/sirupsen/logrus.Entry).Errorf does not support error-wrapping directive %w`.
I suggest for now to leave `%v`.
Thanks

pkg/controlplane/control/dns.go
aviweit: @elevran , done and tested locally.
Can you please review if it is fine in respect to passing them by reference/value?

pkg/controlplane/control/manager.go
aviweit: I think I will set it as resolved.

pkg/controlplane/control/manager.go
aviweit: @elevran , looking into `hosts` plugin, in case it is configured with malformed syntax, coredns pods seem to also fail.


```
$ kubectl --kubeconfig ~/.kube/cl-peer2  get pod   -n kube-system
NAME                                             READY   STATUS             RESTARTS      AGE
coredns-5d78c9869d-klk2t                         0/1     CrashLoopBackOff   5 (61s ago)   4m11s

$ kubectl --kubeconfig ~/.kube/cl-peer2  logs coredns-5d78c9869d-klk2t    -n kube-system
/etc/coredns/custom.hosts:1 - Error during parsing: Unknown directive '1.2.3.4'
```

**coredns configmap:**

```
apiVersion: v1
data:
  Corefile: |
    .:53 {
        import custom.hosts                    < ---- added
        errors
        health {
           lameduck 5s
        }
        ready
        kubernetes cluster.local in-addr.arpa ip6.arpa {
           pods insecure
           fallthrough in-addr.arpa ip6.arpa
           ttl 30
        }
        prometheus :9153
        forward . /etc/resolv.conf {
           max_concurrent 1000
        }
        cache 30
        loop
        reload
        loadbalance
        hosts custom.hosts                    < ---- added
    }
  custom.hosts: |                    < ---- added
    1.2.3.4 mycustom.host
kind: ConfigMap

```

**coredns deployment:**

```
...
      containers:
      - args:
        - -conf
        - /etc/coredns/Corefile
        image: registry.k8s.io/coredns/coredns:v1.10.1
        imagePullPolicy: IfNotPresent
...
        volumeMounts:
        - mountPath: /etc/coredns
          name: config-volume
          readOnly: true

...
      volumes:
      - configMap:
          defaultMode: 420
          items:
          - key: Corefile
            path: Corefile
          - key: custom.hosts                    < ---- added
            path: custom.hosts                    < ---- added
          name: coredns
        name: config-volume


```

config/operator/rbac/role.yaml
aviweit: It doesn't work for me when moving configmap to a namespaced role. I committed the changes under `corednsrewrite-role2` branch derived from this PR branch.

On `AddImport`, ` cl-controlplane` fails with:

```
User "system:serviceaccount:clusterlink-system:cl-controlplane" cannot list resource "configmaps" in API group "" at the cluster scope
E0621 17:56:05.920935       1 reflector.go:150] pkg/mod/k8s.io/client-go@v0.30.2/tools/cache/reflector.go:232: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: configmaps is forbidden: User "system:serviceaccount:clusterlink-system:cl-controlplane" cannot list resource "configmaps" in API group "" at the cluster scope
```

orozery:
@aviweit can you give examples of aliases (those you set in `Import.spec.alias`) for your use-case?

aviweit:
> @aviweit can you give examples of aliases (those you set in `Import.spec.alias`) for your use-case?

@orozery , 

* `vpce-03e2c268057b83e8b-o9f5atr7.lambda.us-east-1.vpce.amazonaws.com`
* `*.vpce-0e1b0638f39353e77-ls3zxnij.s3.us-east-1.vpce.amazonaws.com`

pkg/controlplane/control/manager.go
aviweit: I tried to perform similar to the suggestion in this [link](https://stackoverflow.com/questions/61335318/how-to-restart-a-deployment-in-kubernetes-using-go-client):

1. retrieve `coredns` deployment
2. update it with `restartedAt` annotation
3. call client.Update

It fails to retrieve the deployment with the below error:

```
ERROR  [2024-06-23 15:47:09] failed to configure CoreDNS: no kind is registered for the type v1.Deployment in scheme
"pkg/runtime/scheme.go:100".  component=controlplane.control.manager file=/home/weit/clusterlink/pkg/
controlplane/control/manager.go line=266
```

I can see that operator main() adds all object types to scheme [here](https://github.com/aviweit/clusterlink/blob/corednsrewrite/cmd/cl-operator/main.go#L60) and [here](https://github.com/kubernetes/client-go/blob/master/kubernetes/scheme/register.go#L84).

aviweit:
The PR is set with the discussed changes.

It was tested with kind and AWS EKS clusters.

Thanks.

aviweit:
I have tested this PR with two kind clusters using various `Import` CRs some of them with invalid `alias`:

1. during the rollout (patch) there is always at least one coredns pod in running state; I am able to `curl www.google.com`
2. if the coredns pods fail to start, there is still at least one pod in running state (probably being configured with the old version of config map); if additional `Import`s are being added, configmap is updated, however, the running coredns pod still uses the old good version
4. when deleting the corrupted Import CR(s), rollout is success and coredns pods are available and the other `Import`s that are valid, can be served.

```
echo "
apiVersion: clusterlink.net/v1alpha1
kind: Import
metadata:
  name: nginx-project
  namespace: default
spec:
  port:  80
  alias: www.example.com
  sources:
    - exportName:       nginx-project
      exportNamespace:  default
      peer:             cl-peer1
" | kubectl --kubeconfig ~/.kube/cl-peer2 apply -f -
```

```
echo "
apiVersion: clusterlink.net/v1alpha1
kind: Import
metadata:
  name: nginx-project-wildcard
  namespace: default
spec:
  port:  80
  alias: \"*.wildcard.example.com\"
  sources:
    - exportName:       nginx-project
      exportNamespace:  default
      peer:             cl-peer1
" | kubectl --kubeconfig ~/.kube/cl-peer2 apply -f -
```

```
echo "
apiVersion: clusterlink.net/v1alpha1
kind: Import
metadata:
  name: nginx-project-another
  namespace: default
spec:
  port:  80
  alias: www.example-another.com
  sources:
    - exportName:       nginx-project
      exportNamespace:  default
      peer:             cl-peer1
" | kubectl --kubeconfig ~/.kube/cl-peer2 apply -f -
```

```
echo "
apiVersion: clusterlink.net/v1alpha1
kind: Import
metadata:
  name: nginx-project-no-alias
  namespace: default
spec:
  port:  80
  sources:
    - exportName:       nginx-project
      exportNamespace:  default
      peer:             cl-peer1
" | kubectl --kubeconfig ~/.kube/cl-peer2 apply -f -

```

```
echo "
apiVersion: clusterlink.net/v1alpha1
kind: Import
metadata:
  name: nginx-project-invalid
  namespace: default
spec:
  port:  80
  alias:  $www.example.com
  sources:
    - exportName:       nginx-project
      exportNamespace:  default
      peer:             cl-peer1
" | kubectl --kubeconfig ~/.kube/cl-peer2 apply -f -
```

```
echo "
apiVersion: clusterlink.net/v1alpha1
kind: Import
metadata:
  name: nginx-project-invalid-2
  namespace: default
spec:
  port:  80
  alias:  \"*.www....example-another.com\"
  sources:
    - exportName:       nginx-project
      exportNamespace:  default
      peer:             cl-peer1
" | kubectl --kubeconfig ~/.kube/cl-peer2 apply -f -
```

```
echo "
apiVersion: clusterlink.net/v1alpha1
kind: Import
metadata:
  name: nginx-project-invalid-3
  namespace: default
spec:
  port:  80
  alias:  www www.example-another.com
  sources:
    - exportName:       nginx-project
      exportNamespace:  default
      peer:             cl-peer1
" | kubectl --kubeconfig ~/.kube/cl-peer2 apply -f -

```

I have also tested with two AWS EKS clusters in respect to internal service (e.g. incluster nginx-project) and external ones e.g. AWS s3, AWS lambda.

The policies being used for all tests where "allow all".

config/operator/rbac/role.yaml
elevran, line 1: q: was this file updated to allow getting and changing the CoreDNS deployment objects?
Did we have the permission before? If not, I'd expect new lines in this file.

config/operator/rbac/role.yaml
elevran: That's fine for now, then. Please open an issue to narrow permissions from cluster wide to specific namespace/object later and we'll do it in a follow up PR

pkg/controlplane/control/dns.go
```diff
@@ -0,0 +1,181 @@
...
+
+	if err := mClient.Get(ctx, corednsName, &cm); err != nil {
+		if k8serrors.IsNotFound(err) {
+			logger.Warnf("coredns configmap not found.")
+			return nil
+		} else {
+			return err
+		}
+	}
+	if data, ok := cm.Data["Corefile"]; ok {
```
elevran, line 71: is there a CoreDNS package to parse and edit Corefile instead of manually manipulating string?

pkg/operator/controller/instance_controller.go
```diff
@@ -72,6 +72,7 @@ type InstanceReconciler struct {
 // +kubebuilder:rbac:groups="discovery.k8s.io",resources=endpointslices,verbs=list;get;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=nodes,verbs=list;get;watch
 // +kubebuilder:rbac:groups="",resources=pods,verbs=list;get;watch
+// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;update;watch
 // +kubebuilder:rbac:groups=clusterlink.net,resources=exports;peers;accesspolicies;privilegedaccesspolicies,verbs=list;get;watch
```
elevran, line 76: Need permission to get the Deployment? Is this the latest version of the file?

elevran (approved):

config/operator/rbac/role.yaml
aviweit: These permissions already exist in `roles.yaml` file.

pkg/operator/controller/instance_controller.go
aviweit: It already exists:

```
// +kubebuilder:rbac:groups="apps",resources=deployments,verbs=list;get;watch;create;update;patch;delete
```

config/operator/rbac/role.yaml
aviweit: An issue had been opened: https://github.com/clusterlink-net/clusterlink/issues/660

pkg/operator/controller/instance_controller.go
aviweit: Already removed pod deletion as it is not needed.

pkg/controlplane/control/manager.go
aviweit: I managed.

pkg/controlplane/control/dns.go
aviweit: It seems there is. I found a blog on this:
https://medium.com/jamf-engineering/mastering-coredns-configuration-parsing-the-corefile-with-go-reflection-9180f121a69

elevran:
@aviweit - thank you for this contribution - we're looking for the the best way to make it available in ClusterLink, *safely*. Thus, I'm marking it as `on hold` for now.

The main concern is that `Import`s are namespace scoped but using an alias has cluster-wide impact. This would elevate a user from having only namespace local permissions to affecting workloads across the entire cluster (e.g., the DNS alias could be set to anything, including grabbing service names from another namespace or affecting external services used by workloads in another namespace).

The maintainers will follow up on this work to address the above concern and allow users to experiment with it.
For example, we're considering putting this behind a feature flag (with default set to disabled), adding policies for which aliases are allowed in which namespace, using `pod.Spec.hostAliases` to affects specific Pods in *only* the namespace where the import is defined, etc.

elevran (changes requested):
@orozery @kfirtoledo need to figure out how to enable this functionality while giving administrators control over what can be aliased without negatively affecting entire cluster or "escaping" from the import's namespace

elevran:
@aviweit - thank you for the feature enhancement proposal and PR work that accompanied it.

At this point, we feel that using Import aliases (a namespace scoped construct) to affect CoreDNS configuration (a cluster wide resource, typically under tight administrative control) would not be something we can maintain securely over the long run. Thus, I will close the current PR.

We are exploring alternative ways to support this. One possible approach might be adding a per namespace DNS server to provide aliases (which *could* be based on CoreDNS and the `rewrite`s approach you implemented). This would constrain the effects to the namespace where import aliases are defined and is under full control of namespace users to opt-in (install DNS service and configure workloads to use it).

Thank you again for the PR!

PR #659, opened 2024-06-24
Support multiple controlplane replicas

orozery:
This PR adds an option for deploying clusterlink with multiple controlplane replicas.

### DIFF
--- a/cmd/clusterlink/cmd/deploy/deploy_peer.go
+++ b/cmd/clusterlink/cmd/deploy/deploy_peer.go
@@ -73,7 +73,9 @@ type PeerOptions struct {
 	ContainerRegistry string
 	// Tag represents the tag of the project images.
 	Tag string
-	// Dataplanes is the number of dataplanes to create.
+	// ControlplaneReplicas is the number of controlplanes to create.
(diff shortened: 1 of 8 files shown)


cmd/clusterlink/cmd/deploy/deploy_peer.go
```diff
@@ -185,6 +188,7 @@ func (o *PeerOptions) Run() error {
 		FabricCertificate:       fabricCert,
 		PeerCertificate:         peerCert,
 		CACertificate:           caCert,
+		Controlplanes:           o.ControlplaneReplicas,
```
elevran, line 191: if we maintain backward compatibility by defining that if replicas is not set at all (i.e., the zero value for uint16 is set by default) we treat it as if it was set to `1` - we can save changes across multiple files.
Somewhere we can say:
```go 
   ...
   if controlplaneReplicas == 0 { // not set, use default replica count
      controlplaneReplicas = 1
   }
```

elevran (approved):
Please consider if we can use the unset replica to mean run only 1 instance, to maintain backward compatibility

PR #664, opened 2024-07-03
Add container readiness probes

orozery:
This PR adds container readiness probes to the controlplane and dataplane (both go and envoy) containers.

### DIFF
--- a/cmd/cl-controlplane/app/server.go
+++ b/cmd/cl-controlplane/app/server.go
@@ -15,9 +15,12 @@ package app
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net/http"
 	"os"
 	"path"
+	"syscall"
 
 	"github.com/bombsimon/logrusr/v4"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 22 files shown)


cmd/cl-controlplane/app/server.go
```diff
@@ -39,9 +42,10 @@ import (
 	"github.com/clusterlink-net/clusterlink/pkg/controlplane/xds"
 	"github.com/clusterlink-net/clusterlink/pkg/util/controller"
 	"github.com/clusterlink-net/clusterlink/pkg/util/grpc"
+	utilhttp "github.com/clusterlink-net/clusterlink/pkg/util/http"
 	"github.com/clusterlink-net/clusterlink/pkg/util/log"
 	"github.com/clusterlink-net/clusterlink/pkg/util/runnable"
-	"github.com/clusterlink-net/clusterlink/pkg/util/tls"
+	utiltls "github.com/clusterlink-net/clusterlink/pkg/util/tls"
```
elevran, line 48: q: why was it needed? Is there another tls import? if so, it should have failed before - no?
nit: `tlsutils` instead?

elevran (approved):

PR #665, opened 2024-07-04
More Pod attributes for PDP: SA and labels

zivnevo:
This PR adds more client-Pod attributes for the PDP to consider. We now have:
* `clusterlink/metadata.clientName` - the value of the Pod's label `app` (if it has one)
* `clusterlink/metadata.clientNamespace` - the Pod's namespace
* `clusterlink/metadata.clientServiceAccount` - the Pod's service account
* `client/metadata.<key>` - the value of the Pod's label `<key>` (one such attribute for each label)

Open questions:
* Do we still need `clusterlink/metadata.clientName`? Users can now simply use `client/metadata.app` instead
* How should we handle Pods for which CL has no info (yet)? Deny their request? Provide the PDP with an empty set of attributes?
* E2E testing issue: how to make sure Pod info was read by CL before the Pod can start sending requests? Currently, the solution is `sleep 3`. Any better ideas?


pkg/controlplane/authz/manager.go
```diff
@@ -218,20 +228,36 @@ func (m *Manager) getPodInfoByIP(ip string) *podInfo {
 	return nil
 }
 
+func (m *Manager) getClientAttributes(req *egressAuthorizationRequest) connectivitypdp.WorkloadAttrs {
+	clientAttrs := connectivitypdp.WorkloadAttrs{GatewayNameLabel: m.getPeerName()}
+	podInfo := m.getPodInfoByIP(req.IP)
+	if podInfo == nil {
+		m.logger.Infof("Pod has no info: IP=%v.", req.IP)
+		return clientAttrs // better return an error here?
```
orozery, line 236: Returning an error will simply yield a RST to the client connection.
I think it's better to keep it the way it is (which still allows for allow-all policies to proceed).

pkg/controlplane/authz/manager.go
```diff
@@ -218,20 +228,36 @@ func (m *Manager) getPodInfoByIP(ip string) *podInfo {
...
+	}
+
+	clientAttrs[ServiceNamespaceLabel] = podInfo.namespace // deprecated
+	clientAttrs[ClientSALabel] = podInfo.serviceAccount
+
+	if src, ok := podInfo.labels["app"]; ok {
+		clientAttrs[ServiceNameLabel] = src // deprecated
```
orozery, line 243: What does it take to remove the deprecated attributes?

tests/e2e/k8s/services/httpecho/client.go
```diff
@@ -77,10 +79,22 @@ func GetEchoValue(cluster *util.KindCluster, server *util.Service) (string, erro
...
+		Name:      EchoClientPodName,
 		Namespace: server.Namespace,
 		Image:     "curlimages/curl",
 		Args:      []string{"curl", "-s", "-m", "1", "http://" + server.Name},
 	})
 	return strings.TrimSpace(body), err
 }
+
+// Sleep allows more time for CL to be notified of the new Pod, so CL can retrieve the Pod's info.
+func RunClientInPodWithSleep(cluster *util.KindCluster, server *util.Service) (string, error) {
```
orozery, line 91: I don't think you need this function. If the non-sleep function fails, you will be covered by the `allowRetry=true` for the `AccessService` function.

pkg/controlplane/authz/manager.go
zivnevo: I think they are being used currently in some experiments, so better wait a bit.
They will not appear in the documentation.

pkg/controlplane/authz/manager.go
zivnevo: This means we are not zero-trust by default.
To implement zero-trust, users will have to add a deny policy, blocking connections from unknown pods.
E.g.:
```yaml
apiVersion: clusterlink.net/v1alpha1
kind: PrivilegedAccessPolicy
metadata:
  name: deny-unknown-pods
spec:
  action: deny
  from:
  - workloadSelector:
      matchExpressions:
      - key: clusterlink/metadata.clientNamespace
        operator: Exists 
  to:
  - workloadSelector: {}
```

zivnevo: This is a bit problematic when having policies with negative selectors (using the `NotIn` and `DoesNotExist` operators).
E.g., I have a policy to deny all connections, except from those originating from namespace `foo`.
Connections from IPs for which CL cannot link a Pod, will always be allowed.
```yaml
apiVersion: clusterlink.net/v1alpha1
kind: AccessPolicy
metadata:
  name: deny-except-from-foo
spec:
  action: deny
  from:
  - workloadSelector:
      matchExpressions:
      - key: clusterlink/metadata.clientNamespace
        operator: NotIn
        value: foo
  to:
  - workloadSelector: {}
```

tests/e2e/k8s/services/httpecho/client.go
zivnevo: How does the retry work when we run the client in a Pod?
Aren't we just creating Pods again and again?

orozery: Right, my bad.
So maybe add a flag that removes the `RestartPolicy: v1.RestartPolicyNever`.

pkg/controlplane/authz/manager.go
orozery: Right.
I think the right thing to do is to do a non-cached GET from the API-server to get the pod info in this case.
I did not find a way to do this via the given client. This will be the best option.

orozery: Found this:
https://github.com/kubernetes-sigs/controller-runtime/issues/585#issuecomment-528102351

So there's a non-cached reader, but it's not available to `authz.Manager`. 
Need to propagate it from `cl-controlplane/app/server.go`.

zivnevo: Apparently, even this is not good enough. Pod may start running and send requests, even before its IP is updated in etcd. From what I see, some non-cached `List()` calls return the Pod with its Node IP, but without its Pod IP. It only gets updated later.

orozery: How long is this window of a working pod with etcd not yet updated?
If not long, maybe we can stall a bit to wait for the update.

elevran:
> This PR adds more client-Pod attributes for the PDP to consider. We now have:
> 
> * `clusterlink/metadata.clientName` - the value of the Pod's label `app` (if it has one)
> * `clusterlink/metadata.clientNamespace` - the Pod's namespace
> * `clusterlink/metadata.clientServiceAccount` - the Pod's service account
> * `client/metadata.<key>` - the value of the Pod's label `<key>` (one such attribute for each label)
> 

Is there a need to designate `client` vs `service` in attribute names?
`clusterlink/metadata.clientNamespace` can just be `clusterlink/metadata.namespace`...

> Open questions:
> 
> * Do we still need `clusterlink/metadata.clientName`? Users can now simply use `client/metadata.app` instead

IMO, we do not. I think it is up to the user to set the labels they care about and we should not treat `app` any different than other labels.

> * How should we handle Pods for which CL has no info (yet)? Deny their request? Provide the PDP with an empty set of attributes?

I would propose we deny on lack of sufficient context, to comply with principles of default deny, and let the client retry (e.g., we could consider reset or timeout as the response).
They **could** have properties that would deny the request so proceeding with insufficient knowledge seems to me as the wrong thing security wise.


> * E2E testing issue: how to make sure Pod info was read by CL before the Pod can start sending requests? Currently, the solution is `sleep 3`. Any better ideas?

Tests can explicitly confirm the Pod has IP then wait 1s or so?

pkg/controlplane/authz/manager.go
```diff
@@ -40,6 +40,10 @@ const (
 	// the number of seconds a JWT access token is valid before it expires.
 	jwtExpirySeconds = 5
 
+	ClientNameLabel       = "clusterlink/metadata.clientName"
+	ClientNamespaceLabel  = "clusterlink/metadata.clientNamespace"
+	ClientSALabel         = "clusterlink/metadata.clientServiceAccount"
+	ClientLabelsPrefix    = "client/metadata."
```
elevran, line 46: suggest being explicit that these are labels/tags (i.e., `metadata.label.*`)?
Also, unsure why the designation changes to `client` here. Why separate client from server and do we need a new "grouping" for attributes (besides `clusterlink`)? If so, consider using `k8s` for the relevant attributes (namespace, serviceaccount, labels, ...) instead of clusterlink and reserve that to attributes relating to CL objects (e.g., peer, import, export, ...)

tests/e2e/k8s/services/httpecho/client.go
orozery: > Right, my bad. So maybe add a flag that removes the `RestartPolicy: v1.RestartPolicyNever`.

This won't work.

How about using
```--retry-delay 1 --retry-all-errors```
in the curl pod command line?

Comment (zivnevo) [reply tests/e2e/k8s/services/httpecho/client.go:91]: yes, this works.

File: pkg/controlplane/authz/manager.go (hunk 1 above)
Comment (zivnevo) [reply pkg/controlplane/authz/manager.go]: I changed the implementation: now if the client has no attributes (no pod info) AND the PDP has at least one policy which depends on client attributes, the request will be denied, and the client will have to retry.
Note that even if the PDP has no attribute-dependent policies, attribute-less requests can still be denied, depending on the policies (e.g., DENY policies take precedence, or no ALLOW policies)
This is enforced both on egress and on ingress.

File: pkg/controlplane/authz/manager.go [hunk 5]
```diff
@@ -40,6 +40,10 @@ const (
 	// the number of seconds a JWT access token is valid before it expires.
 	jwtExpirySeconds = 5
 
+	ClientNameLabel       = "clusterlink/metadata.clientName"
```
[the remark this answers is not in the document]
zivnevo: The two feel to me like two different creatures.
A service is a declared entity which remains pretty stable thru the system's life time.
A client just comes out of the blue and requests access. It's true that in many systems the set of clients will also be a stable set, but I can also imagine systems where it's not.
Anyway, these are not very strong feelings, and I'm open to proposals.

orozery (approved):

zivnevo:
After some rethinking, I now believe it's better for client and service to have distinct attribute prefixes. The main reason is that they don't have the exact same set of attributes (e.g., only clients have a service account, only services have a name). Having distinct prefixes, will make it easier to ensure users are not using unavailable attributes in their policies. In addition, distinct prefixes make all attributes share the same pattern - a more consistent scheme.

My current proposal for attribute names (inspired by [K8s well-known labels](https://kubernetes.io/docs/reference/labels-annotations-taints/#applyset-kubernetes-io-additional-namespaces)):
### Client (Pod):
- `client.clusterlink.net/namespace`  - pod namespace
- `client.clusterlink.net/service-account` - pod service account
- `client.clusterlink.net/labels.<label-key>`  - pod labels
### Service (export):
- `export.clusterlink.net/name` - Export name
- `export.clusterlink.net/service-name` - the name of the K8s Service behind the Export (not currently implemented)
- `export.clusterlink.net/namespace` - Export namespace
- `export.clusterlink.net/labels.<label-key>` - Export labels
- `export.clusterlink.net/service-labels.<label-key>` - the labels of the k8s Service behind the Export (not currently implemented)
### Peer:
- `peer.clusterlink.net/name`  - Peer name
- `peer.clusterlink.net/labels.<label-key>` - Peer labels (not currently implemented, to be set when calling `clusterlink deploy peer`)

@elevran , @orozery , @kfirtoledo  - your thoughts?

Issue #291, opened 2024-01-29
Setting peer attributes, to be used in access policies


zivnevo:
Is the Peer CRD the right place for this?

PR #679, opened 2024-07-25
Setting Peer attributes - Phase 1

zivnevo:
This PR allows users to set Peer labels (attributes) from the `clusterlink deploy peer` CLI.
These labels propagate thru `platform.Config`, the `Instance` CR and the control-plane CLI, all the way to the authz manager, where they are added to the "source" attributes.

This is a first step towards solving issue #291 .

**Delayed to phase 2:**
Each Peer should keep track of attributes of other Peers (possibly by slightly abusing the health check mechanism).
This will allow adding Peer attributes to the "destination" attributes, so Peer attributes can be used in both `from` and `to` sections of AccessPolicies.


kfirtoledo (approved):
LGTM

PR #686, opened 2024-08-15
Peers to share their labels with other peers thru heartbeat mechanism

### LINKED ISSUE #291: Setting peer attributes, to be used in access policies

zivnevo:
Fixes #291


kfirtoledo (approved):
LGTM

